PR #6, opened 2017-08-23
Ros Integration

SyllogismRXS:
Ray tracing plugin, bullet physics engine plugin modifications, added transform matrix to State class, fixed Quaternion class bugs.


src/plugins/interaction/BulletCollision/BulletCollision.cpp
```diff
@@ -159,14 +158,13 @@ bool BulletCollision::step_entity_interaction(std::list<sc::EntityPtr> &ents,
...
             // What types of sensors need to be attached to this entity?
-            int num_ray_sensors = 0;
             for (auto &kv : int_to_ent_map[id]->sensors()) {
                 if (kv.second->type() == "Ray") {
                     // Create a publisher for this sensor
-                    // "entity_id/ray_sensor_id/pointcloud"
-                    pcl_pubs_[id][num_ray_sensors] =                        \
-                        create_publisher(std::to_string(id) + "/" +         \
-                                         std::to_string(num_ray_sensors)  + \
+                    // "entity_id/RayTrace0/pointcloud"
```
esquires: ending slashes are not necessary here.

src/parse/MissionParse.cpp
```diff
@@ -422,7 +422,7 @@ bool MissionParse::parse(std::string filename) {
             } else if (nm == "controller" || nm == "sensor" || nm == "sensable"
                        || nm == "autonomy") {
                 nm += "0";
-            }
+            }
```
esquires: cpplint

src/math/State.cpp
```diff
@@ -87,4 +87,31 @@ double State::rel_az(const Eigen::Vector3d &other) {
     return Angles::angle_diff_rad(az, quat_.yaw());
 }
 
+Eigen::Matrix4d State::tf_matrix(bool enable_translate)
```
esquires: It seems like this is doing something very similar to Eigen's [toRotationMatrix](http://eigen.tuxfamily.org/dox/classEigen_1_1RotationBase.html#a7f2232646af917ad28bc6acc228e0f75) and [Transform](http://eigen.tuxfamily.org/dox/classEigen_1_1Transform.html). Can we just use those functions?

include/scrimmage/plugin_manager/Plugin.h
```diff
@@ -85,6 +87,8 @@ class Plugin : public std::enable_shared_from_this<Plugin> {
 
     std::map<std::string, PublisherPtr> pubs_;
     std::map<std::string, SubscriberPtr> subs_;
+
+    StatePtr tf_;
```
esquires: It is not obvious what this variable is. Can we rename it?

PR #7, opened 2017-08-28
Enable/disable setting yaw in DoubleIntegrator, and teleportation fcn

laurathepluralized:


laurathepluralized:
There may be a better way to do this. I'll take a look at it and do another pull request in a few minutes if I get it working less hackily than this pull request.

PR #8, opened 2017-08-28
Quad yaw external control

laurathepluralized:
This modification allows yaw to be set in desired_state_ in autonomy plugin, and includes a parameter in DoubleIntegrator that determines whether the motion model sets the yaw based on velocity or sets it based on whatever is specified in desired_state_->quat().yaw(). Also adds teleportation functionality to DoubleIntegrator.


esquires:
I fixed the merge issues and incorporated it into master. Thanks.

PR #12, opened 2017-09-08
if opencv not found, don't compile plugin dependent on it

laurathepluralized:


SyllogismRXS:
The check failed on the cpplint/cppcheck test, but after the merge, it will pass.

PR #19, opened 2017-10-31
OpenAI Tutorial

bogidude:


esquires:
Thanks for the thorough writeup. This is a nice addition to the documentation. There are a few small things. Can you change the following?

* The build error is due to something in master that has since been fixed. Can you rebase onto the latest master?
* For test_openai.py, can you add some type of assert, rename the file, and put it with the other test_openai.py under ``python/tests``?
* In the tutorial, there are a few references to a scrimmage path on a local machine (e.g., `~/scrimmage/scrimmage/build`). Can you change the path to something generic like ``cd /path/to/scrimmage/build``?

esquires:
An equivalent pull request #22 superceded this one so I am closing it.

PR #20, opened 2017-11-02
Compatible with both Python 2.7 and Python 3

laurathepluralized:


laurathepluralized:
I am working on a few more adjustments to the parameter optimizer separately. This pull request is *only* regarding Py2&3 compatibility.

esquires:
Merged. Thanks.

PR #25, opened 2017-12-06
Entity block XML tag for using variance to select position of all ents in block, including first

laurathepluralized:
`missions/all-ents-variance-demo.xml` is a demo of using this tag


esquires:
Thanks. Can you update the documentation for the xml tags to reflect the change? Also, I don't think we need to add the xml file if the only difference is the new tag. If you want to keep it perhaps we could add a test for that mission.

laurathepluralized:
Sure, I'll update the docs to reflect this. That would probably be better documentation of the functionality than a new mission file :D Should I just git-rm the mission file on my end, or is there a way for me to more cleanly remove it from the history so you don't have to go through and cherry-pick?

esquires:
Thanks. Merged.

PR #26, opened 2018-01-23
Added help menu 

crichardson332:
Added help menu overlay that displays key bindings on the screen, toggled by the 'h' key and defaulted to OFF.


esquires:
Thanks. It looks like there are tabs in this pull request. Can you convert them to spaces? Also, can you update the documentation under ``docs/source/overview/viewer.rst``?

PR #35, opened 2018-03-12
include unordered_map to fix compile err

laurathepluralized:
Didn't see this error on Travis, but I had a local compile error due to unordered_map not being included in Plugin.h. Adding that include fixed it for me.


SyllogismRXS:
Merged. Thanks!

PR #47, opened 2018-04-18
Initialized scores, ensured max_flight_time is initialized early enough

laurathepluralized:


laurathepluralized:
Looks like the build failed on travis due to some docker thing? I'm not sure how to fix that, unfortunately.

esquires:
I think this was just a networking error. Thanks.

PR #49, opened 2018-04-19
Allow for multiple of same network, ent inter, metric plugin

SyllogismRXS:
Eric, please test the param common block feature since this commit changes how the XML files are parsed.


esquires:
This worked fine for me. Thanks for checking.

esquires:
This worked fine for me. Thanks for checking.

PR #51, opened 2018-04-21
add ability to run mission level tests 

esquires:


SyllogismRXS:
I like moving some of the logic from the scrimmage executable's main() into simcontrol.

PR #54, opened 2018-04-25
Add boost_graph dependency

bogidude:
Add boost graph to be installed on initial setup


esquires:
Thanks. I think this error didn't show up in travis because libboost-all-dev is installed for gym.

PR #60, opened 2018-05-13
More realistic ContactBlobCamera sensor plugin

sgarciav:
ContactBlobCamera sensor plugin now considers false positives and false negatives.

- **False negatives**:
User specifies the false negative probability. For each detection, the plugin generates a random number between 0 and 1. If it's less than the set probability, then the detected bounding box is skipped to simulate a false negative.

- **False positives**:
User specifies the false positive probability and the maximum number of false positives (N) any given frame can have. For each detection, the plugin generates N random numbers between 0 and 1. The instances that yield a value less than the set probability will be considered as a false positive. For the current frame, a bounding box will be added centered at random pixel coordinates.


SyllogismRXS:
Thanks!

PR #79, opened 2018-06-11
removed find_dependency call in project-config.cmake that was causing…

crichardson332:
… issues in gtri-uav; moved boost dependency to private scope; added scrimmage-boost target for export


esquires:
The build error was unrelated. Thanks.

PR #85, opened 2018-06-13
path_to_lines function

sgarciav:
Re-added the **path_to_lines** function int the ProtoConversions.cpp class to account for some gtri-uav-scrimmage plugins that make use of it.

### DIFF
--- a/include/scrimmage/proto/ProtoConversions.h
+++ b/include/scrimmage/proto/ProtoConversions.h
@@ -33,6 +33,8 @@
 #ifndef INCLUDE_SCRIMMAGE_PROTO_PROTOCONVERSIONS_H_
 #define INCLUDE_SCRIMMAGE_PROTO_PROTOCONVERSIONS_H_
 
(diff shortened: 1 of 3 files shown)


sgarciav:
Travis is failing to build with the following error:

```
[91m/usr/bin/ld: cannot find -lscrimmage-core
/usr/bin/ld: cannot find -lscrimmage-python
```

I have not seen this error when building on my machine.

esquires:
Adding @crichardson332. Can you comment on the error?

SyllogismRXS:
Sergio, once this travis build builds successfully (https://travis-ci.org/gtri/scrimmage/builds/392339844), try to resubmit your commit.

crichardson332:
Looks like the scrimmage_ros project is failing to link to the new scrimmage-core and scrimmage-python targets that are now part of master. Has this been rebased onto master since merge commit 209200c?

sgarciav:
Thanks guys!

PR #98, opened 2018-07-03
fix derivative term bug in PID

crichardson332:


SyllogismRXS:
Could we just add "prev_error_ = error;" at the end of the step() function instead of adding another member variable? Or does that miss something subtle?

crichardson332:
done

esquires:
Thanks.

Issue #103, opened 2018-07-10
Error: Create an Autonomy Plugin


savasturk:
Hi I'm following this tutorial (http://www.scrimmagesim.org/docs/sphinx/html/tutorials/autonomy-plugin.html ) and run these commands. But when run "make" command I received this error.

~/scrimmage/my-scrimmage-plugins/build$ make
[ 25%] Building CXX object src/plugins/autonomy/ExamplePlugin/CMakeFiles/ExamplePlugin_plugin.dir/ExamplePlugin.cpp.o
In file included from /opt/scrimmage/x86_64-linux-gnu/include/scrimmage/autonomy/Autonomy.h:37:0,
                 from /home/pinar/scrimmage/my-scrimmage-plugins/include/my-scrimmage-plugins/plugins/autonomy/ExamplePlugin/ExamplePlugin.h:35,
                 from /home/pinar/scrimmage/my-scrimmage-plugins/src/plugins/autonomy/ExamplePlugin/ExamplePlugin.cpp:33:
/opt/scrimmage/x86_64-linux-gnu/include/scrimmage/plugin_manager/Plugin.h:36:23: fatal error: Eigen/Dense: No such file or directory
compilation terminated.
src/plugins/autonomy/ExamplePlugin/CMakeFiles/ExamplePlugin_plugin.dir/build.make:62: recipe for target 'src/plugins/autonomy/ExamplePlugin/CMakeFiles/ExamplePlugin_plugin.dir/ExamplePlugin.cpp.o' failed
make[2]: *** [src/plugins/autonomy/ExamplePlugin/CMakeFiles/ExamplePlugin_plugin.dir/ExamplePlugin.cpp.o] Error 1
CMakeFiles/Makefile2:207: recipe for target 'src/plugins/autonomy/ExamplePlugin/CMakeFiles/ExamplePlugin_plugin.dir/all' failed
make[1]: *** [src/plugins/autonomy/ExamplePlugin/CMakeFiles/ExamplePlugin_plugin.dir/all] Error 2
Makefile:149: recipe for target 'all' failed
make: *** [all] Error 2

I use 16.04 LTS Ubuntu dual boot.
pip 10.0.1
Python 3.6.3

How can I solve this error?
Thanks.

esquires:
It looks like the tutorial was out of date so I updated it. When the documentation job completes it will show up on scrimmagesim.org. In the meantime, you can build it locally with:

```[bash]
cd build
[...]

PR #104, opened 2018-07-10
add eigen include dirs to install tree

crichardson332:
Added Eigen include dirs as interface include directories for
scrimmage-core in the install tree. This will fix potential
\#include errors for Eigen when scrimmage has been installed
to a nonstandard location.


esquires:
Thanks. The failure was due to a keyserver error so I added a new issue #105. I built the dockerfile locally and it worked fine. I'll wait to see if this fixes #103 before merging.

esquires:
Thanks

PR #108, opened 2018-07-14
allow users to scale text

esquires:


esquires:
Good suggestion. I just pushed a commit, squashed it, and put it on top of the latest master.

SyllogismRXS:
Thanks, looks good. merged.

PR #121, opened 2018-07-30
Fix sensor shapes not disappearing

jpfeltracco:
Sensor drawn shapes weren't disappearing in the viewer after draw_shape() even with persistent set to false and a small TTL set.

Introduced by https://github.com/gtri/scrimmage/commit/b294986148b42cc6d79c5e8e84702c41de229870 which allowed sensors to display shapes in the viewer.


SyllogismRXS:
Thanks!

PR #124, opened 2018-08-06
updates to run on ubuntu1804

esquires:


esquires:
Yes, the system packages in 18.04 for Protobuf and GRPC are 3.0.0-9 and 1.3.2-1.1, respectively.

https://packages.ubuntu.com/bionic/libprotobuf-dev
https://packages.ubuntu.com/bionic/libgrpc++-dev

The former is behind the ppa and the latter is ahead but I didn't see any api issues running it on 1804 with the different versions.

PR #126, opened 2018-08-14
Fix missing underscore noisy state

jpfeltracco:
Refactor of noisy state sensor left out an underscore causing the xml params of form "pos_noise_0" to not be read properly.


SyllogismRXS:
The Travis build failed, but due to a network problem. It failed during the push of the final docker image to Dockerhub, so I'm going to accept this.

PR #127, opened 2018-08-15
Add an ellipse shape

jpfeltracco:
Couple questions I had.

1. Do we want degrees or radians for the angle field of the ellipse?
2. Do we maybe want to do a quaternion instead? Things like the circle shape currently have no orientation, but there's no reason they couldn't right? Obviously z-axis rotation is pointless, but maybe we could want a sphere not on the xy plane at some point?

### DIFF
--- a/include/scrimmage/plugins/autonomy/ShapeDraw/ShapeDraw.h
+++ b/include/scrimmage/plugins/autonomy/ShapeDraw/ShapeDraw.h
@@ -1,0 +1,56 @@
+/*!
+ * @file
+ *
+ * @section LICENSE
+ *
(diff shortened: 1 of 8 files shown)


esquires:
I am inclined to say radians so that things are kept SI and think your idea for quaternions is a good one.

jpfeltracco:
Okay, updated the ellipse shape to use a quaternion orientation. I also added an autonomy plugin + mission to just draw a few shapes for easy testing. It can be run with `scrimmage missions/shape_draw.xml`

esquires:
Thanks. The error was due to a networking issue. Merged.

PR #128, opened 2018-08-20
Add a shape that draws meshes

jpfeltracco:
Can be tested with `scrimmage missions/shape_draw.xml` A zephyr is being rendered through the shape api.


SyllogismRXS:
The 16.04 dockerfile failed due to a network issue, but this looks good. Thanks.

PR #130, opened 2018-08-22
Updates to External class and ContactBlobCamera plugin

sgarciav:
The three main changes in this merge request:

1. Allow to set radius of agents via the Contact class.
2. Added hooks to the ContactBlobCamera sensor plugin to allow for updating plugin parameters via the External class.
3. Allow the External class to load a single plugin instead of having to load a complete entity.

### DIFF
--- a/include/scrimmage/entity/Contact.h
+++ b/include/scrimmage/entity/Contact.h
@@ -90,6 +90,8 @@ class Contact {
 
     void set_active(bool active);
     bool active();
+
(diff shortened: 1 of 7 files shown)


SyllogismRXS:
The docker push failed at the end of the build, so I'm going to accept this. I am going to update the travis yml file to not run the docker push when using a pull request.

sgarciav:
Thanks!

PR #132, opened 2018-08-24
Added planar boundary in SphereNetwork.

gtcooke94:
There is now an option to add a planar boundary to the sphere network.
A use case example is to model underwater communications not being able
to go beyond the surface of the water to aerial vehicles and vice versa.

### DIFF
--- a/include/scrimmage/plugins/network/SphereNetwork/SphereNetwork.h
+++ b/include/scrimmage/plugins/network/SphereNetwork/SphereNetwork.h
@@ -61,7 +61,7 @@ class SphereNetwork : public scrimmage::Network {
     // The buffer zone around the boundary plane that doesn't directly block
     // comms
     double comms_boundary_epsilon_;
-    bool filter_comms_plane_;
+    bool filter_comms_plane_ = false;
     bool within_planar_boundary(double z1, double z2);
 };
 } // namespace network
--- a/include/scrimmage/plugins/network/SphereNetwork/SphereNetwork.h
+++ b/include/scrimmage/plugins/network/SphereNetwork/SphereNetwork.h
@@ -53,6 +53,16 @@ class SphereNetwork : public scrimmage::Network {
                                             const scrimmage::PluginPtr &sub_plugin) override;
     double range_;
     double prob_transmit_;
+
+    // Attributes of a boundary plane that blocks communication. Example use
+    // would be putting a boundary at a water/air intersection such that
+    // communications wouldn't go between water and the surface.
+    double comms_boundary_altitude_;
+    // The buffer zone around the boundary plane that doesn't directly block
+    // comms
+    double comms_boundary_epsilon_;
+    bool filter_comms_plane_;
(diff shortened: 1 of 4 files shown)


include/scrimmage/plugins/network/SphereNetwork/SphereNetwork.h
```diff
@@ -53,6 +53,16 @@ class SphereNetwork : public scrimmage::Network {
                                             const scrimmage::PluginPtr &sub_plugin) override;
     double range_;
     double prob_transmit_;
+
+    // Attributes of a boundary plane that blocks communication. Example use
+    // would be putting a boundary at a water/air intersection such that
+    // communications wouldn't go between water and the surface.
+    double comms_boundary_altitude_;
```
SyllogismRXS: Can you add appropriate default values for the variables in the header file? (e.g., bool filter_comms_plane_ = false;)

src/plugins/network/SphereNetwork/SphereNetwork.cpp
```diff
@@ -65,6 +65,9 @@ bool SphereNetwork::init(std::map<std::string, std::string> &mission_params,
 
     range_ = std::stod(plugin_params.at("range"));
     prob_transmit_ = std::stod(plugin_params.at("prob_transmit"));
+    filter_comms_plane_ = std::stoi(plugin_params.at("filter_comms_plane"));
```
SyllogismRXS: Can you use the sc::get<T>() calls instead of std::stoi? (e.g., filter_comms_plane_ = sc::get<bool>("filter_comms_plane", plugin_params, filter_comms_plane_);)

include/scrimmage/plugins/network/SphereNetwork/SphereNetwork.xml
```diff
@@ -4,6 +4,9 @@
   <library>SphereNetwork_plugin</library>
   <range>100</range>
   <prob_transmit>1.0</prob_transmit>
+  <filter_comms_plane>0</filter_comms_plane>
```
SyllogismRXS: use "true" or "false" in filter_comms_plane.

gtcooke94:
Thanks, I'll make these changes.
When done should I just push up another commit to this branch or squash them? Also, should I leave it where it is or rebase off of master again?

SyllogismRXS:
You can push up another commit. Don't worry about squashing or rebasing. I think there won't be any conflicts.

include/scrimmage/plugins/network/SphereNetwork/SphereNetwork.h
gtcooke94: @SyllogismRXS Do you want me to add default values in the header for the doubles? i.e. for comms_boundary_altitiude_ and comms_boundary_epsilon_. If they aren't in the params file I have them default to quietNaN right now when they are being read in.

SyllogismRXS:
Thanks!

PR #137, opened 2018-09-10
add macOS support

crichardson332:
This pull request is being made to run the CI tests and create discussion about merging in macOS support. This has been tested on two macOS systems, where scrimmage can be built from source natively and run. A Homebrew package has also been set up that enables easy installation of all dependencies, which will be hosted [here](https://github.com/crichardson332/homebrew-crich_brews), along with instructions on installing dependencies and building/running scrimmage.

Main changes in this commit:
- add no-return-type-c-linkage flags to remove warnings form clang about extern "C" block in RegisterPlugin macro
- add const keywords to waypoint generator
- disabled everything using linux joystick in macOS build
- add some missing overrides
- add missing condition_variable.h include in SimControl
- add missing signal.h includes


crichardson332:
The failure here is due to the test_cpplint.py failing. It complains that <condition_variable> is an unapproved c++11 header, which you can see [here](https://chromium-cpp.appspot.com/) is apparently banned by Google. This is strange though considering condition_variable is used all over scrimmage's code, even without the header included...we also use several other headers that are supposedly banned on that same page.

SyllogismRXS:
There are several binary files that were modified in this commit. Was this by accident? If so, can you revert the changes on the binary files and submit a new commit.

crichardson332:
I was originally getting [this warning](https://stackoverflow.com/questions/22745076/libpng-warning-iccp-known-incorrect-srgb-profile) whenever running scrimmage with the GUI enabled, and was able to fix it by running pngcrush on the wing*.png files. I just rebased on the latest merge and things seem to be broken now, so I'll have to take a look at it and then reopen the pull request.

share/cmake/scrimmage/GenerateSetEnv.cmake
```diff
@@ -2,11 +2,11 @@ include(CMakeParseArguments)
 
 function(GenerateSetEnv)
   set(options)
-  set(oneValueArgs SETUP_LOCAL_CONFIG_DIR LOCAL_CONFIG_DIR SETENV_IN_FILE)
+  set(oneValueArgs SETUP_HOME_CONFIG LOCAL_CONFIG_DIR SETENV_IN_FILE)
```
SyllogismRXS: Can you change the SETUP_HOME_CONFIG variable name back to SETUP_LOCAL_CONFIG_DIR? The SETUP_LOCAL_CONFIG_DIR option is directly related to the LOCAL_CONFIG_DIR, the LOCAL_CONFIG doens't necessarily have to be located in the user's $ENV{HOME}, but it defaults to HOME if the LOCAL_CONFIG_DIR isn't specified.

PR #138, opened 2018-09-10
Connect controller variable IO in serial, example energy plugin

SyllogismRXS:


src/entity/Entity.cpp
```diff
@@ -215,26 +215,55 @@ bool Entity::init(AttributeMap &overrides,
...
+                                                   motion_model_->vars());
+        if (controller == nullptr) {
+            std::cout << "Failed to open controller plugin: " << controller_name << std::endl;
+            return false;
+        }
 
-    if (controller_ == nullptr) {
-        std::cout << "Failed to open controller plugin: " << it->second << std::endl;
-        return false;
+        if (!verify_io_connection(controller->vars(), motion_model_->vars())) {
```
esquires: Having this check for every controller's connection to the motion model seems like it may raise some false positives. For instance, if we want the serial connection controller1 -> controller2 -> motion, then this is going to check that controller1 ->motion is connected correctly.

scripts/templates/motion/PluginTemplate.h
```diff
@@ -44,22 +41,19 @@ namespace scrimmage {
...
-    virtual void model(const vector_t &x , vector_t &dxdt , double t);
-
-    class Controller : public scrimmage::Controller {
-     public:
-        virtual Eigen::Vector2d &u() = 0;
-    };
  protected:
-    double length_;
-    bool enable_gravity_;
+    int speed_idx_ = 0;
```
esquires: can we default to using ``uint8_t``? It seems a bit safer for indexing.

src/entity/Entity.cpp
SyllogismRXS: Yes, I agree. I moved that verify_io_connection to where the last controller is connected to the motion model.

scripts/templates/motion/PluginTemplate.h
SyllogismRXS: Yes, sounds good.

PR #144, opened 2018-09-24
Override parameters from external, plugin tags for external

SyllogismRXS:
Eric,
  I have updated the external class and plugin creation functions to use the tags attribute in all plugins to enable or disable loading of that plugin. Also, we can now override plugin parameters from a process that calls the External class' create_entity function (e.g., override scrimmage plugins with ROS parameters at initialization time).  Please take a look and let me know what you think. This commit will fail to build until I also update the scrimmage_ros package, so don't accept this merge request, but provide any feedback that you have. I will need to update the documentation for using scrimmage within a ROS nodelet as well.

### DIFF
--- a/include/scrimmage/plugin_manager/PluginManager.h
+++ b/include/scrimmage/plugin_manager/PluginManager.h
@@ -116,14 +116,15 @@ class PluginManager {
             bool valid_tag = false;
             auto it_tags = overrides.find("tags");
             if (it_tags != overrides.end()) {
-                // Do any of the plugin tags match the tags for this plugin?
-                std::set<std::string> tags = str2set<std::string>(it_tags->second, ", ");
-                for (const std::string &tag: tags) {
-                    if (plugin_tags.find(tag) != plugin_tags.end()) {
-                        valid_tag = true;
-                        break;
-                    }
-                }
+                // Break up the plugin's tags attribute into a set of strings
+                auto tags = str2container<std::set<std::string>>(it_tags->second, ", ");
+                // If the plugin_tags contains any tag in tags, this plugin
+                // should be loaded
+                valid_tag = std::any_of(tags.begin(), tags.end(),
+                                        [&](const std::string &tag) {
+                                            return (plugin_tags.find(tag) !=
+                                                    plugin_tags.end());
+                                        });
             }
             if (not valid_tag) {
                 status.status = PluginStatus<T>::invalid_tag;
--- a/include/scrimmage/plugin_manager/PluginManager.h
+++ b/include/scrimmage/plugin_manager/PluginManager.h
@@ -34,12 +34,23 @@
 #define INCLUDE_SCRIMMAGE_PLUGIN_MANAGER_PLUGINMANAGER_H_
 
 #include <scrimmage/fwd_decl.h>
+#include <scrimmage/parse/ConfigParse.h>
+#include <scrimmage/parse/ParseUtils.h>
+#include <scrimmage/common/FileSearch.h>
 
+#include <iostream>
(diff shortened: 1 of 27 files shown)


src/simcontrol/SimControl.cpp
```diff
@@ -264,9 +264,9 @@ bool SimControl::init() {
     info.id_to_ent_map = id_to_ent_map_;
 
     networks_ = std::make_shared<NetworkMap>();
-    if (!create_networks(info, *networks_)) return false;
-    if (!create_metrics(info, contacts_, metrics_)) return false;
-    if (!create_ent_inters(info, contacts_, shapes_[0], ent_inters_)) return false;
+    if (!create_networks(info, *networks_, std::set<std::string>{}, [](std::map<std::string, std::string>&){})) return false;
```
esquires: I think the last two arguments would be easier to read if it was just 

```[c++]
     if (!create_networks(info, *networks_, {}, {})) return false; 
```

missions/uuv-ex1.xml
```diff
@@ -37,11 +36,12 @@
   <origin_length>10</origin_length>
 
   <entity_interaction>SimpleCollision</entity_interaction>
+  <metrics tags="REAL">SimpleCollisionMetrics</metrics>
 
   <enable_screenshots min_period="1.0" start="8.3" end="15.3">false</enable_screenshots>
 
-  <network>GlobalNetwork</network>
-  <network>LocalNetwork</network>
+  <network tags="REAL">GlobalNetwork</network>
```
esquires: While we are updating the tags, can we unify this convention with what is done for entity? In other words, can we update it from 

```[xml]
  <entity>
    <name>uuv_entity</name>
  </entity>
``` 

to 

```[xml]
<entity tags="uuv_entity">
</entity>
 ```

File: src/common/Utilities.cpp [hunk 3]
```diff
@@ -49,6 +49,19 @@ namespace fs = boost::filesystem;
 
 namespace scrimmage {
 
+std::string exec_command(const char* cmd) {
+    std::array<char, 128> buffer;
+    std::string result;
+    std::shared_ptr<FILE> pipe(popen(cmd, "r"), pclose);
+    if (!pipe) throw std::runtime_error("popen() failed!");
```
esquires: I think ``std::unique_ptr`` would be more appropriate here.

src/common/Utilities.cpp
```diff
@@ -49,6 +49,19 @@ namespace fs = boost::filesystem;
 
 namespace scrimmage {
 
+std::string exec_command(const char* cmd) {
```
esquires: Since we aren't actually using this command in scrimmage, it seems more appropriate to put it in scrimmage_ros

include/scrimmage/plugin_manager/PluginManager.h
```diff
@@ -55,6 +66,14 @@ class PluginInfo {
     void * handle = nullptr;
 };
 
+template <class T>
+class PluginStatus {
```
esquires: Good idea. This will make failures more informative

src/common/Utilities.cpp
SyllogismRXS: I agree. I'm going to move this function out of scrimmage core.

src/simcontrol/SimControl.cpp
SyllogismRXS: Agreed

missions/uuv-ex1.xml
SyllogismRXS: Yes, I'll use entity tags instead.

PR #146, opened 2018-09-24
add custom keypress using ';' key

crichardson332:


esquires:
short of the linting error, this looks fine to me.

PR #147, opened 2018-09-28
Working scrimmage param server

SyllogismRXS:


include/scrimmage/common/ParameterServer.h
```diff
@@ -0,0 +1,138 @@
...
+        // Search for the parameter name in the server
+        auto it_name = params_.find(name);
+        if (it_name != params_.end()) {
+            // The param name exists, does the type exist?
+            auto it_type = it_name->second.find(typeid(T).name());
+            if (it_type != it_name->second.end()) {
+                // The param name/type exists in the list, iterate over the
+                // list of parameters and call their set_value functions
+                for (ParameterBasePtr param : it_type->second) {
+                    auto param_cast = std::dynamic_pointer_cast<Parameter<T>>(param);
```
esquires: This seems inefficient to check every time ``set_param`` is called. I think we can move this type check to ``register_param`` so it is only done once.

include/scrimmage/common/ParameterServer.h
```diff
@@ -0,0 +1,138 @@
...
+        }
+        // Could not find the appropriate param name/type/plugin-owner
+        return false;
+    }
+
+ protected:
+    // Key 1: parameter name (string)
+    // Key 2: parameter type (as a string)
+    // Value: ParameterBase shared_ptr
+    std::map<std::string, std::map<std::string, std::list<ParameterBasePtr>>> params_;
```
esquires: unordered_map may be more efficient here. Also, I think instead of ``list``, we should be using ``std::set``. This will get rid of the ``std::remove_if``` logic above.

include/scrimmage/common/ParameterServer.h
```diff
@@ -0,0 +1,138 @@
...
+            // The parameter name doesn't exist it, create it.
+            params_[name][typeid(T).name()] = std::list<ParameterBasePtr> {param};
+        } else {
+            // Does this type exist for this parameter name?
+            auto it_type = it_name->second.find(typeid(T).name());
+            if (it_type != it_name->second.end()) {
+                // This param name / type combo exists
+                // Has this specific plugin already registered this parameter?
+                for (const ParameterBasePtr &ptr : it_type->second) {
+                    if (ptr->owner() == owner) {
```
esquires: using ``std::set`` in ``params_`` will mean we don't need this logic.

include/scrimmage/common/ParameterServer.h
```diff
@@ -0,0 +1,138 @@
...
+        return false;
+    }
+
+ protected:
+    // Key 1: parameter name (string)
+    // Key 2: parameter type (as a string)
+    // Value: ParameterBase shared_ptr
+    std::map<std::string, std::map<std::string, std::list<ParameterBasePtr>>> params_;
+};
+typedef std::shared_ptr<ParameterServer> ParameterServerPtr;
```
esquires: can we use ``using`` instead of ``typedef`` to be more inline with ``c++11`` style?

include/scrimmage/entity/External.h
```diff
@@ -36,6 +36,7 @@
 #include <scrimmage/common/RTree.h>
 #include <scrimmage/common/Time.h>
 #include <scrimmage/common/DelayedTask.h>
+#include <scrimmage/common/ParameterServer.h>
```
esquires: we can use a forward declaration here.

src/common/ParameterServer.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace scrimmage {
+void ParameterServer::unregister_params(PluginPtr owner) {
+    // For all parameters, remove all parameters owned by this plugin
+    for (auto &kv1 : params_) {
+        for (auto &kv2 : kv1.second) {
+            kv2.second.erase(std::remove_if(kv2.second.begin(), kv2.second.end(),
+                                            [&](ParameterBasePtr ptr) {
+                                                return ptr->owner() == owner; }),
+                             kv2.second.end());
```
esquires: See comment about hard to read before. I think if we use ``std::set`` in ``params_`` this will also become simpler.

src/plugin_manager/Plugin.cpp
```diff
@@ -50,10 +51,12 @@ Plugin::Plugin() : name_("Plugin"), parent_(std::make_shared<Entity>()),
...
                    id_to_team_map_(std::make_shared<std::unordered_map<int, int>>()),
                    id_to_ent_map_(std::make_shared<std::unordered_map<int, EntityPtr>>()),
-                   time_(std::make_shared<const Time>()) {
+                   time_(std::make_shared<const Time>()),
+                   param_server_(std::make_shared<ParameterServer>()) {
 }
 
-Plugin::~Plugin() {}
+Plugin::~Plugin() {
+}
```
esquires: do we need this extra line?

src/plugins/autonomy/ArduPilot/ArduPilot.cpp
```diff
@@ -146,7 +146,7 @@ void ArduPilot::start_receive() {
                     ba::placeholders::bytes_transferred));
 }
 
-void ArduPilot::close(double t) {
+void ArduPilot::close(const double &t) {
```
esquires: Is there a need to add a reference here? This will break the api for downstream plugins. In particular, if they are not using ``override`` they won't know about the change and ``close`` would just not be called.

src/plugins/interaction/SimpleCollision/SimpleCollision.cpp
```diff
@@ -44,6 +46,7 @@
 
 #include <limits>
 #include <memory>
+#include <iostream>
```
esquires: are all of these includes necessary? It looks like the file otherwise didn't change.

src/plugin_manager/Plugin.cpp
SyllogismRXS: changed during a debug, i'll change back

src/plugins/autonomy/ArduPilot/ArduPilot.cpp
SyllogismRXS: Ok. We have to figure out when we will update the step(), init(), and close() functions to better follow const correctness.

src/plugins/interaction/SimpleCollision/SimpleCollision.cpp
SyllogismRXS: left over during debug, removed.

include/scrimmage/common/ParameterServer.h
SyllogismRXS: ok

include/scrimmage/entity/External.h
SyllogismRXS: ok

include/scrimmage/common/ParameterServer.h
esquires: I thought about this a bit more and realized we do need the dynamic cast so please disregard.

SyllogismRXS: You are right. I wrote this check before I changed params_ to use have a second map that is keyed based on a string representation of the type (typeid(T).name()). This guarantees that the dynamic_pointer_cast will succeed later when called.

include/scrimmage/common/ParameterServer.h
```diff
@@ -0,0 +1,138 @@
...
+        auto it_name = params_.find(name);
+        if (it_name != params_.end()) {
+            auto it_type = it_name->second.find(typeid(T).name());
+            if (it_type != it_name->second.end()) {
+                // The parameter name/type exists, let's find the specific
+                // plugin
+                auto it_param = std::remove_if(it_type->second.begin(),
+                                               it_type->second.end(),
+                                               [&](ParameterBasePtr ptr) {
+                                                   return ptr->owner() == owner; });
```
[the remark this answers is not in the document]
SyllogismRXS: Updated. Wrote a common inline function to remove_if_owner.

include/scrimmage/common/ParameterServer.h
SyllogismRXS: Updated.

include/scrimmage/common/ParameterServer.h
SyllogismRXS: Updated.

src/common/ParameterServer.cpp
SyllogismRXS: Updated.

PR #149, opened 2018-10-02
Demo contracted graph, load it into GraphInteraction

laurathepluralized:
This fixes a problem with the old version of GraphInteraction that would treat the edges in a contracted graph file as nodes. Nodes and edges are correctly differentiated now, and loading the graph files is done by just changing the data xml file for the location and setting the correct lat long for that location; the details about which graph file to load for that location are collected by sc::parse_autonomy_data.


SyllogismRXS:
I am a little worried about these large pycgrc files. Do these need to be in the repository? Can the user easily generate these files?

laurathepluralized:
As we discussed in person earlier, I swapped out the old pycgrc* files for much smaller (just a couple of city blocks) pycgr and pycgrc files so that we can include them for demonstrating graph drawing and using stuff from the data directory in missions. If they need to be even smaller (maybe one block of a single street?), let me know. I also converted the README.md that was in GraphInteraction's include dir into a chapter of the SCRIMMAGE tutorials and expanded it somewhat. What do you think? 
Thanks!

laurathepluralized:
Hey, I may wind up redoing some of this for a more standard graph file format soon, so it may be better to hold off on merging this after all. Sorry for any confusion!

SyllogismRXS:
Ok. I'll hold off on merging this. I'm going to close this merge request, so that I don't get confused. Feel to open a new one later (or maybe you can reopen this closed one, not sure).

PR #153, opened 2018-10-09
Use the correct variable name for home directory setup in GenerateSetEnv

jpfeltracco:
A mismatch between the template name of a parameter for GenerateSetEnv and the actual parameter name causes no more setenv files to get written to ~/.scrimmage in projects.


esquires:
Looks good. Thanks.

PR #154, opened 2018-10-10
Sim timing

dmagree:
Moved loop_wait call to the end of the loop, so we are waiting only for the extra time in the loop. also changed the timer interface a little


SyllogismRXS:
It looks like the build failed due to a network problem. Can you rebase the branch into one or two commits, so that I can easily view the changes? Since there are some changes to the core simulator logic, I want to make sure there aren't any unintended consequences. Thanks.

SyllogismRXS:
Looks good. Thanks.

PR #156, opened 2018-10-10
GraphInteraction now visualizes OSM data read from graphml files

laurathepluralized:


src/plugins/interaction/GraphInteraction/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ add_library(${LIBRARY_NAME} SHARED
   ${SRCS}
 )
 
+include_directories(${Boost_INCLUDE_DIRS})
```
esquires: Can we use `target_include_directories` instead?

src/plugins/interaction/GraphInteraction/GraphInteraction.cpp
```diff
@@ -42,19 +43,29 @@
 #include <scrimmage/proto/ProtoConversions.h>
 #include <scrimmage/msgs/Graph.pb.h>
 
-
+#include <typeinfo>
 #include <memory>
 #include <limits>
 #include <iostream>
 #include <fstream>
+#include <deque>
```
esquires: is this used anywhere?

include/scrimmage/plugins/interaction/GraphInteraction/GraphInteraction.h
```diff
@@ -41,6 +41,12 @@
 #include <map>
 #include <list>
 #include <string>
+#include <vector>
+#include <deque>
```
esquires: is this used anywhere? It  looks like it should be removed. We can also forward declare Entity, Message, and Publisher.

src/plugins/interaction/GraphInteraction/GraphInteraction.cpp
```diff
@@ -77,112 +88,122 @@ std::ifstream& GotoLine(std::ifstream& file, unsigned int num) {
...
-        pub_graph_->publish(graph_msg);
+    std::pair<boost::adjacency_list<>::vertex_iterator,
+        boost::adjacency_list<>::vertex_iterator> vs = boost::vertices(g_);
+
+    std::map<uint64_t, Eigen::Vector3d> nodes;
+    std::map<uint64_t, uint64_t> boost_vert_to_osmid;
+
+    auto graph_msg = std::make_shared<sc::Message<sm::Graph>>();
+    graph_msg->data.set_id(id_);
+    for (; vs.first != vs.second; ++vs.first) {
```
esquires: can we just initialize this variable in the loop? In other words, `auto vs = boost::vertices(g_); vs.first != vs.second; ++vs.first`.

src/plugins/interaction/GraphInteraction/GraphInteraction.cpp
```diff
@@ -77,112 +88,122 @@ std::ifstream& GotoLine(std::ifstream& file, unsigned int num) {
...
+        parent_->projection()->Forward(latitude, longitude, 0.0, x, y, z);
+        nodes[this_osmid] = Eigen::Vector3d(x, y, z);
+        auto node_ptr = graph_msg->data.add_nodes();
+        node_ptr->set_id(this_osmid);
+        sc::set(node_ptr->mutable_point(), nodes[this_osmid]);
+    }
 
-        // Visualize the Nodes
+    auto es = boost::edges(g_);
+    for (; es.first != es.second; ++es.first) {
```
esquires: can we initialize this variable in the loop?

src/plugins/interaction/GraphInteraction/CMakeLists.txt
laurathepluralized: Sure, though please double-check my use of it in upcoming push; not quite sure whether my use of it covers both the build and install tree parts correctly. Thanks!

src/plugins/interaction/GraphInteraction/GraphInteraction.cpp
laurathepluralized: Good call. Removed in upcoming push. Thanks!

src/plugins/interaction/GraphInteraction/GraphInteraction.cpp
laurathepluralized: Yes, good idea. Thanks!

src/plugins/interaction/GraphInteraction/GraphInteraction.cpp
laurathepluralized: Yes, good idea. Thanks!

src/plugins/interaction/GraphInteraction/CMakeLists.txt
```diff
@@ -12,7 +12,8 @@ add_library(${LIBRARY_NAME} SHARED
   ${SRCS}
 )
 
-include_directories(${Boost_INCLUDE_DIRS})
+target_include_directories(scrimmage-boost
+  INTERFACE ${Boost_INCLUDE_DIRS})
```
esquires: This library needs boost so `INTERFACE` should be `PUBLIC`

laurathepluralized: I think I finally got it. Still a bit confused about target_include_directories and the many many things that can go into them (e.g. all of https://cmake.org/cmake/help/v3.0/command/target_include_directories.html), but that was a good learning experience. Thank you!

laurathepluralized:
So apparently my last attempt didn't work; I guess I just had some cache on my machine helping it compile. I'm guessing I need to delve back into the details of the `INSTALL_INTERFACE` and `BUILD_INTERFACE`, though my attempts at putting various combinations of `${Boost_LIBRARIES}`, `${Boost_INCLUDE_DIRS}`, and `scrimmage-boost` in either of those have all failed. @esquires, can you clarify how to go about setting this up properly?

PR #163, opened 2018-10-21
cmake cleanup

### LINKED ISSUE #150: target_compile_definitions for pybind
Plugin projects need something like this in `CMakeLists.txt`:

```[cmake]
if (SCRIMMAGE_ENABLE_PYTHON_BINDINGS)
  add_definitions(-DENABLE_PYTHON_BINDINGS=1)
else()
  add_definitions(-DENABLE_PYTHON_BINDINGS=0)
endif()
```

We can remove that need by using `target_compile_definitions` for our targets

esquires:
* use target_include_directories instead of include_directories
* use target_compile_definition for ENABLE_PYTHON_BINDINGS

Also resolves #150


SyllogismRXS:
Thanks for updating the CMake project. This looks good when it passes.

PR #175, opened 2018-11-12
Drawing avoidance vector for AvoidEntityMS

gtcooke94:
Optionally draw the Avoidance vector in red when using AvoidEntityMS from MotorSchemas


esquires:
Looks good. Thanks.

PR #179, opened 2018-11-27
switch to rapidxml included with boost

crichardson332:
This commit switches to using the rapidxml that is packaged with boost in the property_tree library. Since we already depend on boost headers, this removes a dependency; also, it allows rapidxml to be used easily on macOS. Presently, there is no good way to install rapidxml on macOS (we've been manually copying the headers for every install). I was prompted to find a better solution by a user who's [trying to build and use scrimmage on mac.](https://github.com/crichardson332/homebrew-crich_brews/issues/1) 

What do you guys think?

### DIFF
--- a/ci/dockerfiles/centos6
+++ b/ci/dockerfiles/centos6
@@ -180,23 +180,6 @@ RUN source /etc/profile.d/rvm.sh && cd /root/installs/geographiclib \
     && rpm -ivh /root/installs/geographiclib/*.rpm
 
 ###############################################################################
-# RapidXML Package
-###############################################################################
-WORKDIR /root/3rd-party
-RUN git clone https://github.com/dwd/rapidxml.git && cd rapidxml \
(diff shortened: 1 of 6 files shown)


crichardson332:
Looks like this change breaks something, so I need to look into it a bit more.

esquires:
I am somewhat concerned about using namespace detail in boost although it seems like we need to make things easier to use on a mac. Can we just add a target to cmake that will download the files similar to what we do under 3rd-party/CMakeLists.txt so we can know we are using the mainline version of rapidxml? Users on a mac would just have an extra build step

```[bash]
#onetime items
cd 3rd-party
mkdir build
cd build
cmake .. 
make rapidxml
make install
# then normal build steps
```

crichardson332:
Alright I just created a homebrew package for rapidxml hosted on my repo so users can brew install it. Now it's just part of scrimmage dependencies so there are no extra steps.

PR #196, opened 2018-12-10
Add Experiment Features

jpfeltracco:
- Add symlink to latest experiment in experiments folder
- Add ability to do a grid search over parameters (with optional count field in ranges file)
- Add ability to repeat a given set of params (can set in run_experiments.py cmdline interface)
- Add tqdm progress bar. This also provides estimated completion times for experiments.


SyllogismRXS:
Thanks!

PR #198, opened 2018-12-13
Removed unnecessary link to Boundary_plugin in the CMakeLists.txt of several plugins

jharrison:
The following CMakeLists.txt files unneccesarily link to Boundary_plugin:
- plugins/autonomy/BoundaryDefense
- plugins/autonomy/Straight
- plugins/autonomy/TakeFlag
- plugins/interaction/CaptureInBoundaryInteraction
- plugins/interaction/EnforceBoundaryInteraction
- plugins/interaction/FlagCaptureInteraction

### DIFF
--- a/src/plugins/interaction/CaptureInBoundaryInteraction/CMakeLists.txt
+++ b/src/plugins/interaction/CaptureInBoundaryInteraction/CMakeLists.txt
@@ -14,7 +14,6 @@ ADD_LIBRARY(${LIBRARY_NAME} SHARED
 
 TARGET_LINK_LIBRARIES(${LIBRARY_NAME}
     scrimmage-core
-  Boundary_plugin
   )
 
 SET (_soversion ${LIB_MAJOR}.${LIB_MINOR}.${LIB_RELEASE})
--- a/src/plugins/interaction/EnforceBoundaryInteraction/CMakeLists.txt
+++ b/src/plugins/interaction/EnforceBoundaryInteraction/CMakeLists.txt
@@ -14,7 +14,6 @@ ADD_LIBRARY(${LIBRARY_NAME} SHARED
 
 TARGET_LINK_LIBRARIES(${LIBRARY_NAME}
     scrimmage-core
-  Boundary_plugin
   )
 
 SET (_soversion ${LIB_MAJOR}.${LIB_MINOR}.${LIB_RELEASE})
--- a/src/plugins/interaction/FlagCaptureInteraction/CMakeLists.txt
+++ b/src/plugins/interaction/FlagCaptureInteraction/CMakeLists.txt
@@ -14,7 +14,6 @@ ADD_LIBRARY(${LIBRARY_NAME} SHARED
 
 TARGET_LINK_LIBRARIES(${LIBRARY_NAME}
     scrimmage-core
-  Boundary_plugin
   )
 
 SET (_soversion ${LIB_MAJOR}.${LIB_MINOR}.${LIB_RELEASE})
--- a/src/plugins/autonomy/BoundaryDefense/CMakeLists.txt
+++ b/src/plugins/autonomy/BoundaryDefense/CMakeLists.txt
@@ -18,7 +18,6 @@ target_compile_options(${LIBRARY_NAME}
 
 target_link_libraries(${LIBRARY_NAME}
   scrimmage-core
-  Boundary_plugin
 )
 
 set(_soversion ${LIB_MAJOR}.${LIB_MINOR}.${LIB_RELEASE})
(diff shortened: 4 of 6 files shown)


esquires:
Hi @jharrison. Thanks for the pull request. That error comes up when bibtex isn't installed (we don't add it to the Docker build to reduce the time it takes to test) but it shouldn't cause a build to fail.

It looks like there is a linker error in one of the tests:

```[bash]
/root/scrimmage/scrimmage/build/bin/test_collisions: symbol lookup error: /root/scrimmage/scrimmage/build/plugin_libs/libStraight_plugin.so: undefined symbol: _ZN9scrimmage11interaction8Boundary13make_boundaryERKN15scrimmage_proto5ShapeE
```

src/plugins/interaction/FlagCaptureInteraction/CMakeLists.txt
```diff
@@ -14,7 +14,6 @@ ADD_LIBRARY(${LIBRARY_NAME} SHARED
 
 TARGET_LINK_LIBRARIES(${LIBRARY_NAME}
     scrimmage-core
-  Boundary_plugin
```
esquires: The boundary plugin is used on line 76 of FlagCaptureInteraction.cpp

src/plugins/interaction/EnforceBoundaryInteraction/CMakeLists.txt
```diff
@@ -14,7 +14,6 @@ ADD_LIBRARY(${LIBRARY_NAME} SHARED
 
 TARGET_LINK_LIBRARIES(${LIBRARY_NAME}
     scrimmage-core
-  Boundary_plugin
```
esquires: The boundary is used on line 78 of EnforceBoundaryInteraction.cpp

src/plugins/interaction/CaptureInBoundaryInteraction/CMakeLists.txt
```diff
@@ -14,7 +14,6 @@ ADD_LIBRARY(${LIBRARY_NAME} SHARED
 
 TARGET_LINK_LIBRARIES(${LIBRARY_NAME}
     scrimmage-core
-  Boundary_plugin
```
esquires: The boundary is used on line 82 of CaptureInBoundaryInteraction.cpp

esquires:
I think most of those plugins need that link. SimControl might be the exception though.

jharrison:
In [another thread](https://github.com/crichardson332/homebrew-crich_brews/issues/3#issuecomment-442570120), @crichardson332 suggested that these weren't needed because Boundary doesn't have a standalone utility class, so I tried removing the link and found that they still compile and build fine on both my mac and my ubuntu 16.04 machine.

crichardson332:
The Boundary plugin has a make_boundary() function that allows dynamic creation of boundary objects, based on input Shapes. This requires explicitly linking to the Boundary plugin for any plugin that uses that functionality. 

In general, any plugin that uses another plugin's API directly, including any auxiliary class that's compiled with that plugin, must explicitly link to that plugin in CMake. Most of the time this is not the case, but there are a few instances of it in scrimmage core.

esquires:
It sounds like from the other thread that the problem is a downstream build. Can it just do the following?

```[cmake]
target_link_libraries(myplugin
    PUBLIC
    Boundary_plugin)
```

jharrison:
I see what you're saying, but straight.xml compiles, links, and runs fine after removing the `Boundary_plugin` link from plugins/autonomy/Straight/CMakeLists.txt. Based on what you're saying, this should fail because Straight.cpp calls `sci::Boundary::make_boundary(msg->data)`.

esquires:
The test that failed in Jenkins failed because of a linker error. To verify on my system, I applied the patch of this pull request and ran this through gdb to confirm that the test fails on the ``make_boundary`` call. It also fails when running the ``straight.xml`` outside of a test harness (not using gdb). 

I am not sure how it is running on your system without the ``target_link_libraries`` command but I think it is safe to say that it is needed for it to work on other systems.

Is it a problem to link to this library? It seems like the easiest solution is to just add add a ``target_link_libraries`` call since the definitions are in a ``.cpp`` file.

crichardson332:
It's not guaranteed to fail after removing the explicit link from Straight's cmake list file, because transitive dependencies can "accidentally" provide links to required libraries and make things build and run. But if you're using a class's API, you must link to the library that class is compiled into. Even if it accidentally worked once on one system, it is not robust and isn't going to be portable or cross platform. That's why we have the Travis CI server and automatic clean builds, to find pesky bugs like this.

jharrison:
I see. Thanks for the explanation. I guess I'll just delete this pull request then.

PR #200, opened 2018-12-19
Patch segfaulting when losing entities on sphere network

jpfeltracco:
At least in the case when entities collide and are set inactive, pub/sub plugins in the pubsub map in SimControl have their parent entity closed, but they stay in the pubsub map. This causes a null pointer deref when something tries to refer to the parent object of the one of these publisher or subscribers.


esquires:
Thanks.

I'll make a ticket to fix ``SimControl::run_remove_inactive`` so plugins don't have to think about this detail.

PR #203, opened 2018-12-20
Allow 0 opacity on shapes, fix cuboid rotation

jpfeltracco:
Lets a user "hide" an active shape by setting the opacity to 0.0 instead of automatically changing 0.0 -> 1.0.
Also fixes a cuboid rotation issue where each subsequent rotation command was getting added to the previous orientation, causing the shape to spin when called in a loop.

Use case of invisible shapes: Say we're drawing a line between our entity and every other entity in a large simulation. Other entities are constantly dying and new ones are spawning. Rather than removing the shape actors over and over for the lines, each simulation loop we can simply default each line in a pool to being invisible, then allocate the ones we're actually using by setting opacity back to a non-zero value.

If others are interested, I could generalize some of my pool code to get it into scrimmage base as well. Here's the basic idea though:

```c++
class ShapePool {
 public:
    scrimmage_proto::ShapePtr allocate() {
        if (num_allocated_ == shapes_.size()) {
            shapes_.push_back(std::make_shared<scrimmage_proto::Shape>());
        }
        num_allocated_++;
        return shapes_[num_allocated_ - 1];
    }

    void reset() {
        for (auto sh : shapes_) {
            sh->set_opacity(0.0);
        }
        num_allocated_ = 0;
    }

    std::vector<scrimmage_proto::ShapePtr>& get_shapes() { return shapes_; }

 protected:
    std::vector<scrimmage_proto::ShapePtr> shapes_;
    std::size_t num_allocated_ = 0;
};

```

Usage:
```c++
// in step function for a plugin
lines_pool_.reset();

for (...) {
    auto shape_ptr = lines_pool_.allocate();
    shape_ptr->set_opacity(1.0);
    sc::set(shape_ptr->mutable_line()->mutable_start(), /* pos */)
    sc::set(shape_ptr->mutable_line()->mutable_end(), /* pos */)
}

for (auto& shape : lines_pool_.get_shapes()) {
    draw_shape(shape);
}
```


SyllogismRXS:
One reason I set the opacity to 1.0 if the opacity is very small (or zero) is because I didn't want the user to have to specify the opacity in order to view the shape. I think the default value for opacity should be 1.0, unless the user specifies it. Does this pull request still support an opacity of 1.0 as default? Also, it looks like protobuf3 dropped support for default values in message fields.

jpfeltracco:
Ah, makes sense. What if we have some sort of message generation function that sets up some sane defaults like opacity = 1.0?

That function could also set the shape hash during creation. That might be conceptually simpler than having the draw function setting the hash as a side effect too.

Or we could add a `bool opacity_set` type field, similar to the `hash_set` field to prevent people from accidentally sending an opacity of 0.

jpfeltracco:
Okay, I've cut the scope of this PR to just fix that cuboid rotation issue and to clean up the ShapeDraw test plugin a little bit.

PR #219, opened 2019-01-16
Added various error checks in MissionParse and External.

JamesMichaelMatthews:
Added checks for xml parsing errors in MissionParse. However, note that sometimes it hard crashes, sometimes not.
Fixed sub_cb return that was getting bad data from a bad lambda.
Added error checking in the External step function, to catch the case where scrimmage was not setup properly. Also found a missing mutex unlock.

### DIFF
--- a/include/scrimmage/entity/External.h
+++ b/include/scrimmage/entity/External.h
@@ -178,7 +178,8 @@ class External {
         boost::optional<std::list<NetworkDevicePtr>> subs =
             pubsub_->find_subs(network_name, topic_name);
 
-        auto no_subs_msg = [&]() {
+        auto no_subs_msg = [=]() { //capture by copy, used in returned function.
+            //Note: Find_subs already prints this information. Keep it here or there?
             std::cout << "No subscribers exist in SCRIMMAGE" << std::endl;
             std::cout << "Network name: " << network_name << std::endl;
             std::cout << "Topic name: " << topic_name << std::endl;
@@ -346,7 +347,7 @@ class External {
 #endif
 
  protected:
-    void call_update_contacts(double t);
+    bool call_update_contacts(double t); //false on error
     void update_time(double t);
 
     std::list<InterfacePtr> outgoing_interfaces_;
--- a/include/scrimmage/entity/External.h
+++ b/include/scrimmage/entity/External.h
@@ -178,22 +178,10 @@ class External {
         boost::optional<std::list<NetworkDevicePtr>> subs =
             pubsub_->find_subs(network_name, topic_name);
 
-        auto no_subs_msg = [=]() { //capture by copy, used in returned function.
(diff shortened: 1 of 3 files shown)


src/parse/MissionParse.cpp
```diff
@@ -84,7 +84,19 @@ bool MissionParse::parse(const std::string &filename) {
...
+    //doc.parse requires a null terminated string
+    std::vector<char> contentvec(content.size() + 1); //allocation done here
+    contentvec.assign(content.begin(), content.end()); //copy
+    contentvec.push_back('\0'); //shouldn't reallocate
+    try {
+      //Note: This parse function can hard fail (seg fault, no exception) on
+      //      badly formatted xml data. Sometimes it'll except, sometimes not.
+      doc.parse<0>(contentvec.data());
+    } catch (...) {
+      cout << "scrimmage::MissionParse::parse: Exception during rapidxml::xml_document<>.parse<>()." << endl;
```
esquires: Can we throw an error here instead of using cout?

include/scrimmage/entity/External.h
```diff
@@ -178,7 +178,8 @@ class External {
         boost::optional<std::list<NetworkDevicePtr>> subs =
             pubsub_->find_subs(network_name, topic_name);
 
-        auto no_subs_msg = [&]() {
+        auto no_subs_msg = [=]() { //capture by copy, used in returned function.
+            //Note: Find_subs already prints this information. Keep it here or there?
```
esquires: If it is already printed in find_subs, can we just make this function empty?

src/entity/External.cpp
```diff
@@ -255,7 +255,7 @@ bool External::step(double t) {
     last_t_ = t;
     mutex.unlock();
 
-    this->call_update_contacts(t);
+    if(!this->call_update_contacts(t)) return false;
```
esquires: This is probably a cpplint error.

src/parse/MissionParse.cpp
JamesMichaelMatthews: It is already. It returns false, that's the error mechanism used here.

include/scrimmage/entity/External.h
JamesMichaelMatthews: Idk. I would, but it wasn't like that before, so you decide. Maybe you'd prefer to remove it from the other location, how'd you want to design it. You wrote this didn't you?

src/entity/External.cpp
JamesMichaelMatthews: Looks good to me. What's the problem?

JamesMichaelMatthews:
I had to rebase again. Initially I didn't realize I rebased onto origin/master but didn't fetch that remote. Fixed the conflicts.

esquires:
Thanks!

PR #223, opened 2019-01-28
Delay comms

squishbug:
Add option for delay in simulated network comms using comm_delay argument in network definition (see missions/delay-test.xml. Stochastic delay can be implemented in user-defined network plugin but may cause significant slowdown of the simulation.

### DIFF
--- a/include/scrimmage/pubsub/Network.h
+++ b/include/scrimmage/pubsub/Network.h
@@ -66,10 +66,18 @@ class Network : public Plugin {
     inline virtual void set_mission_parse(MissionParsePtr mp)
     { mp_ = mp; }
 
+    // Get/set delay values
+    double comm_delay() {return comm_delay_;}
+    void set_comm_delay(double del) {comm_delay_ = del;}
+
  protected:
     RTreePtr rtree_;
     RandomPtr random_;
     MissionParsePtr mp_;
+
+    // Delay handling
+    double comm_delay_ = -1;
+    bool is_stochastic_delay_ = false;
 
     // Key 1: Publisher Entity ID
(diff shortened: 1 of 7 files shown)


include/scrimmage/pubsub/Network.h
```diff
@@ -66,11 +66,19 @@ class Network : public Plugin {
...
+    double comm_delay() {return comm_delay_;}
+    void set_comm_delay(double del) {comm_delay_ = del;}
+
  protected:
     RTreePtr rtree_;
     RandomPtr random_;
     MissionParsePtr mp_;
 
+    // Delay handling
+    double comm_delay_ = 0;
```
SyllogismRXS: Can you default comm_delay_ to -1. It isn't always safe to check for comm_delay_ == 0 due to the computer precision. Instead, check for comm_delay_ < 0. A negative delay doesn't make sense, so we can disable the comm_delay feature by making the delay negative.

src/pubsub/Network.cpp
```diff
@@ -149,21 +157,45 @@ bool Network::step(std::map<std::string, std::list<NetworkDevicePtr>> &pubs,
...
-                            msg->time = time_->t();
-                            sub->add_msg(msg);
 
-                            if (monitor_all_subs_) {
-                                // Accumulate received message counts on all topics
-                                it_all_sub->second += 1;
-                            }
+                            double msg_delay = get_transmission_delay();
+
+                            if (msg_delay == 0) {
```
SyllogismRXS: Check for msg_delay < 0 instead.

SyllogismRXS:
Thanks for the new communications delay feature!

PR #229, opened 2019-01-30
Add grpc non-learning mode documentation and tests

bogidude:
Also fixes various bugs in the grpc non-learning mode


esquires:
Looks good. Thanks.

PR #233, opened 2019-02-01
clean up const functions in State

crichardson332:
Hey guys, I wanted to suggest the idea of refactoring the getters/setters for the State class (and potentially any other class where it's relevant). The current API has 3 different methods for accessing pos/vel/etc - 1 getter, 1 setter, and 1 that can do both. This has caused me confusion and led to some very time consuming bugs related to erroneously mixing return by copy and return by reference. I've also had a few instances where I made extensive use of 'pos()' as a getter, which worked fine for a time until I ran into a situation where my object was now const. Then every instance of 'pos()' had to be changed to 'pos_const()'.

I believe these headaches can all be avoided by replacing the 3 current function calls with just 1, utilizing const overloads that the compiler will resolve for you. This way the API is simpler and you never have to think about which version of the function to use, and the compiler will use the const version whenever it can and the mutable version when it needs to.

What do you guys think?

### DIFF
--- a/include/scrimmage/math/State.h
+++ b/include/scrimmage/math/State.h
@@ -55,15 +55,19 @@ class State {
     Eigen::Vector3d &pos();
     Eigen::Vector3d &vel();
     Eigen::Vector3d &ang_vel();
+    const Eigen::Vector3d &pos() const;
+    const Eigen::Vector3d &vel() const;
+    const Eigen::Vector3d &ang_vel() const;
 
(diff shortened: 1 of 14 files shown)


SyllogismRXS:
I would like to make all of our API const correct. We just need to figure out how much this will affect everyone's plugins.

crichardson332:
I updated this pull request to keep the current API for backwards compatibility, so it shouldn't break any existing plugins.

PR #242, opened 2019-02-06
add loop_rate api to plugins

crichardson332:
What do you guys think about adding the ability to control step rates for plugins?

### DIFF
--- a/docs/source/tutorials/autonomy-plugin.rst
+++ b/docs/source/tutorials/autonomy-plugin.rst
@@ -37,7 +37,7 @@ Let's take a look at the header file
 
 .. code-block:: c++
    :linenos:
-                
+
     #ifndef INCLUDE_MY_SCRIMMAGE_PLUGINS_PLUGINS_AUTONOMY_MYFOLLOWBEHAVIOR_MYFOLLOWBEHAVIOR_H_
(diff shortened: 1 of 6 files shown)


SyllogismRXS:
I like the idea of the user being able to change the loop rate for various plugins from the main mission file XML. I don't think the user should have to parse / modify the loop_rate in their own init() function though. It should be handled automatically by the mission parser / sim_control. We have the concept of the "motion_multiplier," which I would like to extend to all plugins, but it is confusing to people in its current implementation.

crichardson332:
I originally tried to do it through an xml tag in the plugin, but the problem is that then gets conflated with user-controlled params. A user might create a param with the same name unknowingly, and accidentally change the loop rate of their plugin and have no idea why it isn’t stepping every time. 

There’s also the benefit of the user having more control if it’s an API instead of a param, where it could be changed dynamically throughout the sim with any arbitrary trigger. 

Do you have any ideas of how to get around the ambiguity of user params vs sim control variables?

~ Christopher Richardson

SyllogismRXS:
We already have the problem of ambiguity of user and sim params for a couple of features (for example, the "tags" attribute). I think this is fine for now as long as we document it in the tutorials. We will eventually be updating our mission file format (maybe not using XML in the future) and we will come up with a way to distinguish between these types of parameters. 

I think we can still allow the plugin user to modify the loop_rate through the C++ API and handle the most obvious use case for them behind the scenes.

PR #250, opened 2019-02-11
Fix bugs in nonlearning grpc mode

bogidude:
Bugs fixed include nonexistent symbol kill_grpc_server() being called when built with ENABLE_GRPC=0 and accidentally killing all process ids when the variable python_cmd_ is not initialized.

### DIFF
--- a/src/plugins/autonomy/ScrimmageOpenAIAutonomy/ScrimmageOpenAIAutonomy.cpp
+++ b/src/plugins/autonomy/ScrimmageOpenAIAutonomy/ScrimmageOpenAIAutonomy.cpp
@@ -182,6 +182,7 @@ bool ScrimmageOpenAIAutonomy::step_autonomy(double t, double /*dt*/) {
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/plugins/autonomy/ScrimmageOpenAIAutonomy/ScrimmageOpenAIAutonomy.cpp
```diff
@@ -182,18 +182,24 @@ bool ScrimmageOpenAIAutonomy::step_autonomy(double t, double /*dt*/) {
...
             kill_grpc_server();
         }
     }
+    #endif
     return step_helper();
 }
 
 #if ENABLE_GRPC
 void ScrimmageOpenAIAutonomy::kill_grpc_server() {
+    // Check if python_cmd_ is too short
```
esquires: can you add a comment on the motivation for this check?

esquires:
Thanks!

PR #261, opened 2019-02-20
build when grpc is not found

esquires:


laurathepluralized:
I was just about to file an issue about scrimmage not building if it had trouble finding my grpc installation, but this seems to have fixed it. Thank you!

PR #268, opened 2019-03-05
Use sc::get() to parse params in Unicycle model

SyllogismRXS:


SyllogismRXS:
I added default values to the Unicycle motion model in Unicycle.h and Unicycle.xml. When some of the new parameters were not defined in the mission file, I received a map out of range error. I also changed the params.at("") calls to using our sc::get<T>() parsing helper.

PR #288, opened 2019-04-08
Cleanup external's use of time

jpfeltracco:
External requires a time as an input to its step function, but it also refers to rostime internally.

This makes the interface more consistent, letting external only use the time passed in through step.

### DIFF
--- a/include/scrimmage/entity/External.h
+++ b/include/scrimmage/entity/External.h
@@ -76,6 +76,8 @@ class External {
                        const std::string &plugin_tags_str,
                        int entity_id,
                        int max_entities,
+                       double init_time,
+                       double init_dt,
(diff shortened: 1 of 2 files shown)


include/scrimmage/entity/External.h
```diff
@@ -187,12 +188,18 @@ class External {
                 return;
             }
 
+            if (!step_called_at_least_once_) {
```
esquires: Can we add an initial time to the init function so we don't run into this situation?

jpfeltracco:
Sure! Just updated.

jpfeltracco:
Sure, I made a PR yesterday here: https://github.com/SyllogismRXS/scrimmage_ros/pull/3

PR #292, opened 2019-04-16
initial working dispatcher prototype

crichardson332:
I added an initial prototype for a dispatcher program. I called it 'scrimpy' to avoid conflicts with existing executables, but that can be changed later if we want this to be the main entry point into scrimmage. 

Running 'pip install -e .' inside the scrimmage/python directory should install it into ~/.local, and right now all arguments are just forwarded to the existing programs. Tab completion can be enabled by sourcing 'scrimmage/python/completion/scrimmage-argcomplete.bash'. For some reason pip isn't installing that script for me, even though it's specified in the setup file (I've read that data_files can be problematic with pip python packages). I also plan to extend tab completion by adding custom completers that can list all the mission files, plugins, etc. on the scrimmage path that are visible to the parser. 

Let me know what you guys think.

### DIFF
--- a/python/completion/scrimmage-argcomplete.bash
+++ b/python/completion/scrimmage-argcomplete.bash
@@ -1,0 +1,7 @@
+# GNU Lesser General Public License v3.0
+
+if type register-python-argcomplete3 > /dev/null 2>&1; then
+  eval "$(register-python-argcomplete3 ros2)"
+elif type register-python-argcomplete > /dev/null 2>&1; then
+  eval "$(register-python-argcomplete ros2)"
+fi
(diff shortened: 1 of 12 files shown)


SyllogismRXS:
Well, I hate and love the name, "scrimpy". I'll try it out soon and let you know. Thanks for starting this.

crichardson332:
Btw, the tab completion is fast when it's run as a script, but it gets slow when it's installed by pip. I looked into it and it's because it's installed as an 'entry_point' in setup.py, which makes it into an executable script. But when you do that python adds an import statement into the script that's very slow - `from pkg_resources import load_entry_point`. This causes tab completion to be slow for the installed executable because it has to run this import every time you hit tab. 

I haven't figured out how to fix this yet, I'll have to look into it some more.

PR #294, opened 2019-04-17
add polyline shape

crichardson332:


SyllogismRXS:
Can you add the VTK viewer function to draw the polyline?

crichardson332:
Done.

PR #299, opened 2019-04-26
Add dynamic OpenAI interface state space

bogidude:
Allows the OpenAI interface to handle agents to come in and out of the simulation

### DIFF
--- a/python/scrimmage/bindings/src/py_openai_env.cpp
+++ b/python/scrimmage/bindings/src/py_openai_env.cpp
@@ -96,11 +96,13 @@ ScrimmageOpenAIEnv::ScrimmageOpenAIEnv(
         bool enable_gui,
         bool combine_actors,
         bool global_sensor,
+        bool static_obs_space,
         double timestep) :
     spec(py::none()),
     metadata(py::dict()),
     mission_file_(mission_file),
-    enable_gui_(enable_gui) {
+    enable_gui_(enable_gui),
+    static_obs_space_(static_obs_space) {
 
     observations_.set_global_sensor(global_sensor);
     observations_.set_combine_actors(combine_actors);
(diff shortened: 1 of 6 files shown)


python/scrimmage/bindings/src/py_openai_env.cpp
```diff
@@ -355,8 +358,34 @@ void ScrimmageOpenAIEnv::seed(pybind11::object _seed) {
     seed_ = _seed.cast<int>();
 }
 
-pybind11::tuple ScrimmageOpenAIEnv::step(pybind11::object action) {
+void ScrimmageOpenAIEnv::filter_ext_ctrl_vec(){
```
esquires: can we just use ``erase`` with ``remove_if``? This would look something like:

```[c++]
auto no_parent = [&](auto &p){return p->parent() == nullptr;};
std::erase(std::remove_if(actions_.ext_ctrl_vec().begin(), actions_.ext_ctrl_vec().end(), no_parent), actions_.ext_ctrl_vec().end());
```

python/scrimmage/bindings/src/py_openai_env.cpp
```diff
@@ -355,8 +358,34 @@ void ScrimmageOpenAIEnv::seed(pybind11::object _seed) {
...
+        auto parent_ptr = autonomy_ptr->parent();
+        if (parent_ptr == nullptr) {
+            it = actions_.ext_ctrl_vec().erase(it);
+        } else {
+            it++;
+        }
+    }
+}
 
+void ScrimmageOpenAIEnv::filter_ext_sensor_vec(){
```
esquires: same comment as above. Can we use erase with remove_if?

src/plugins/autonomy/ScrimmageOpenAIAutonomy/OpenAIObservations.cpp
```diff
@@ -108,6 +108,17 @@ pybind11::object OpenAIObservations::update_observation(size_t num_entities) {
             observation_space.attr("spaces").cast<py::list>();
         py::list observation_list = observation.cast<py::list>();
 
+        // Get python len() function
+        py::object len_ = py::module::import("builtins").attr("len");
+
+        // Number of entities has shrunk
+        while (len_(observation_list).cast<size_t>() > ext_sensor_vec_.size()) {
```
esquires: pybind11 has a built-in len function that we can use. See [here](https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L1331)

esquires:
Thanks!

PR #300, opened 2019-05-01
Initial Shape API

SyllogismRXS:


SyllogismRXS:
Several of our users have their own wrapper functions around our protobuf-based shape API. I think we should provide the a shape wrapper API to scrimmage developers to reduce code across plugins and repos.

crichardson332:
Looks good, although did you mean for the first function to be "make_state" as opposed to "make_arrow"?

Also, do you think we could add a general "make_shape" function that creates a shape with sensible defaults, and maybe takes in a color? 99% of the time I make a shape, I want persistent(true) and opacity to be 1.0, but protobuf defaults those to off.

PR #302, opened 2019-05-07
WIP: Message logging

crichardson332:
I wanted to know what you guys thought about this. One thing some of us think would be really nice is if we had a way to log messages being published into the scrimmage system in some kind of binary format that could be played back later, sort of like rosbags. The main problem with this is that scrimmage messages are just shared pointers to c++ classes and can't be readily serialized and logged.

However, I think boost serialization might offer us a solution. It supports all native c++ data types for serialization, including STL containers, and allows you to write pretty minimal code to expose your own custom classes to its API. If you attempt to serialize a class that doesn't have the necessary hooks written, it will fail at build time. I was playing around with it and was able to write a plugin that publishes a std::vector of scrimmage::State objects, which is then serialized into XML and logged in a file. This could enable us to log all messages being passed through the system and be able to recreate them later, which would be especially helpful when running scrimmage plugins inside of a larger context.

Also, we could utilize this to create a "rostopic echo" like function that could read this log file during the simulation and print its data to the screen, which is another thing I know several people have been wishing for.

Let me know what you guys think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ tags*
 **.swp
 **.DS_Store
 **.ccls*
+**compile_commands.json
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,7 +196,7 @@ if (${VTK_FOUND})
         set(OPENGL_MODULE vtkRenderingOpenGL2)
     endif()
 
-    set(VTK_MODULES 
+    set(VTK_MODULES
         vtkFiltersSources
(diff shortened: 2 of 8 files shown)


crichardson332:
Oh I forgot to mention, if you want to test this out, just pull this branch and run  `missions/test/test_pubsub.xml`. It should create a `test_boost_serial.xml` file that shows the logged messages.

PR #303, opened 2019-05-07
allow to override yaw value in SingleIntegrator plugin

sgarciav:
The yaw value in the SingleIntegration motion plugin is currently computed as a function of the velocity vector. However, there are uses cases where the user might want to set a specific heading that is not in the direction of movement. This commit introduces the ``override_heading`` parameter. If set to true, then the heading of the vehicle will be set to whatever value is published to the ``OverrideHeading`` topic.

### DIFF
--- a/include/scrimmage/plugins/autonomy/MotorSchemas/MotorSchemas.h
+++ b/include/scrimmage/plugins/autonomy/MotorSchemas/MotorSchemas.h
@@ -51,6 +51,7 @@ class MotorSchemas : public scrimmage::Autonomy {
 
  protected:
     bool show_shapes_;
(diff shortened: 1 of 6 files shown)


sgarciav:
Updated the solution. Instead of using the ``OverrideHeading`` topic to publish the heading value directly to the SingleIntegrator motion plugin, the solution leverages the VariableIO class. The SingleIntegrator plugin still has the ``override_heading`` parameter which, if set to true, will set the aircraft's heading to whatever value is passed via the VariableIO call.

In this manner, the controller plugin will also know of the value provided to the motion model.

esquires:
Thanks!

PR #305, opened 2019-05-08
add sleep() for pause efficiency

crichardson332:
- link to opencv libraries properly in exported target


crichardson332:
This commit makes the simulator significantly more efficient in terms of cpu usage while paused.

esquires:
This seems like it is going to sleep even when the simulation is not paused. Can we add an if statement so it is not slowed down when not paused?

crichardson332:
Done.

esquires:
Thanks!

PR #306, opened 2019-05-08
Add hil support

dmagree:


include/scrimmage/plugins/autonomy/ArduPilot/ArduPilot.xml
```diff
@@ -6,9 +6,12 @@
   <!-- Transmit Socket: State information from SCRIMMAGE to ArduPilot -->
   <to_ardupilot_ip>127.0.0.1</to_ardupilot_ip>
   <to_ardupilot_port>5003</to_ardupilot_port>
+  <from_ardupilot_port>5002</from_ardupilot_port>
 
   <!-- Receive Socket: PWM commands from ArduPilot -->
-  <from_ardupilot_port>5002</from_ardupilot_port>
+<!--   <to_ardupilot_port>5501</to_ardupilot_port>
```
SyllogismRXS: Can you remove commented out xml if it is not needed? If it's used in a specific mode, maybe add a comment here.

src/plugins/autonomy/ArduPilot/ArduPilot.cpp
```diff
@@ -110,9 +110,12 @@ void ArduPilot::init(std::map<std::string, std::string> &params) {
         }
     }
 
+    mavproxy_mode_ = sc::get<bool>("mavproxy_mode", params, false);
+
     // Get parameters for transmit socket (to ardupilot)
     to_ardupilot_ip_ = sc::get<std::string>("to_ardupilot_ip", params, "127.0.0.1");
     to_ardupilot_port_ = sc::get<std::string>("to_ardupilot_port", params, "5003");
+    printf("ArduPilot: sending to udp:%s:%s\n", to_ardupilot_ip_.c_str(), to_ardupilot_port_.c_str());
```
SyllogismRXS: Can you use std::cout instead of printf? We try to limit the amount of text printed to the screen unless it is a warning or error. Can you wrap your couts in if-statements (enable_debug), where enable_debug is a plugin parameter?

src/plugins/autonomy/ArduPilot/ArduPilot.cpp
```diff
@@ -180,7 +184,7 @@ bool ArduPilot::step_autonomy(double t, double dt) {
 
 void ArduPilot::handle_receive(const boost::system::error_code& error,
                                std::size_t num_bytes) {
-#if 0
+#if 1
```
SyllogismRXS: Should this #if evaluate to true on the "final" version of the ArduPilot plugin? Maybe wrap this in the enable_debug parameter. If you want different levels of debug, you could use a debug_level parameter instead, where debug_level is an integer.

SyllogismRXS:
Thanks!

PR #307, opened 2019-05-08
Add port checking for GRPC server in nonlearning mode

bogidude:
Fix broken tests for OpenAI


esquires:
This was merged in #316

PR #313, opened 2019-05-14
fixed CSV logging documentation

sgarciav:
There were missing commas in the append block.


SyllogismRXS:
Thanks for the fixes!

PR #320, opened 2019-05-30
Pose

crichardson332:
I made a pose class that can be used for algorithms that need timestamped 3d points with orientation, but don't need the complications of reference frames. Our waypoint class could eventually utilize this class and add the necessary geographic coordinate information and handle re-linearization and whatnot, but for algorithms that are frame-agnostic they can use this. It's modeled after the ROS geometry messages.

### DIFF
--- a/include/scrimmage/common/Pose.h
+++ b/include/scrimmage/common/Pose.h
@@ -42,6 +42,7 @@ namespace scrimmage {
 class Pose {
  public:
     Pose();
+    virtual ~Pose();
     explicit Pose(const Eigen::Vector3d &pos);
     explicit Pose(const scrimmage::Quaternion &quat);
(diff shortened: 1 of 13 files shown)


SyllogismRXS:
What are the benefits of this Pose class over just using the State class that already exists? I see that you may want a time stamp attached to the Pose/State. Any other benefits?

crichardson332:
Other than needing the timestamp, the main benefit would just be memory/space efficiency whenever someone only needs pose and nothing else.

We could update the State class to contain a Pose object instead of keeping its own pos/quat, and maintain the same api but just forward the pos() and quat() functions to the underlying Pose class.

crichardson332:
I just made a commit that shows the corresponding changes to State, let me know what you think.

PR #341, opened 2019-07-05
Improve 3rd-party/readme

flaviut:
Break it up into paragraphs so it is easier to scan over & use code
formatting where appropriate.


SyllogismRXS:
Thanks!

PR #343, opened 2019-07-08
Get 3rd-party deps building on my machine

flaviut:
Hi, I've been working on getting the third-party dependencies building on my machine.

I've made two changes so far:

- add `-Wno-error` on grpc build so that new warnings that are found in newer compiler versions don't break the build.
- Specify `PKG_CONFIG_PATH` so that grpc looks in the right place for the protobuf library files. Before making this change, grpc's build process found my system's protobuf2 library and failed to compile.

I haven't been able to test this patch fully either on my machine or on launchpad, due to more issues with building grpc. I'll update this patch as I solve the issues.


flaviut:
I have tested this on my machine, with all the relevent directories deleted, and the HEAD of this branch builds successfully, builds scrimmage successfully, and runs scrimmage successfully.

SyllogismRXS:
Thanks.

PR #345, opened 2019-07-24
Fix arducopter for ardupilot simulations

avelte3:


src/plugins/motion/Multirotor/Multirotor.cpp
```diff
@@ -307,6 +307,9 @@ void Multirotor::model(const vector_t &x , vector_t &dxdt , double t) {
 
     for (unsigned int i = 0; i < rotors_.size(); i++) {
         rotor_thrust(i) =  c_T_ * omega_sq(i);
+        if (omega[i] == wmin_) { //if minimum throttle, servo is off
```
SyllogismRXS: Since two doubles are being compared here, you may want to check for less than or equal, where equal is a something like: std::abs(omega[i] - wmin_) < std::numeric_limits<double>::epsilon()

PR #355, opened 2019-09-03
Nps features

shaun-d-anderson:
Core Updates for NPS Mission sets


src/plugins/controller/UnicyclePID/UnicyclePID.cpp
```diff
@@ -93,33 +94,39 @@ void UnicyclePID::init(std::map<std::string, std::string> &params) {
     if (!speed_pid_.init(params["speed_pid"], false)) {
         std::cout << "Failed to set speed PID" << std::endl;
     }
+
+    parent_->state() = std::make_shared<State>();
+    *(parent_->state()) = *(parent_->state_truth());
```
SyllogismRXS: The controller shouldn't be acting on the state's "truth," but a possibly noisy version of the truth since this representative of a real system. The sensor plugin interface handles setting the controller plugin's "state_" variable to be the "truth" or "noisy" versions depending on whether the NoisyState plugin was used or not.

src/plugins/controller/UnicyclePID/UnicyclePID.cpp
```diff
@@ -93,33 +94,39 @@ void UnicyclePID::init(std::map<std::string, std::string> &params) {
...
 }
 
 bool UnicyclePID::step(double t, double dt) {
+    // Make a copy of the current state
+    StateWithCovariance ns(*(parent_->state()));
+
     heading_pid_.set_setpoint(vars_.input(desired_heading_idx_));
-    vars_.output(turn_rate_idx_, heading_pid_.step(time_->dt(), state_->quat().yaw()));
```
SyllogismRXS: I think we should still be using "state_" here, as well as the other places in this file.

src/plugins/autonomy/MotorSchemas/MotorSchemas.cpp
```diff
@@ -75,6 +75,16 @@ void MotorSchemas::init(std::map<std::string, std::string> &params) {
...
+    
+    // Get Max Current Speed and adjust autonomy to be able to go 110% of Max Current
+    // ("GlobalNetwork", "Force_Field")
+    // auto msg_force = std::make_shared<sc::Message<Eigen::Vector3d>>();
+    // sc::sensor::SideScanSonarSensorType SSS_msg_;
+
+    auto callback_force = [&] (auto &msg) {
+            Force_ = msg->data;
+        };
+    subscribe<Eigen::Vector3d>( "GlobalNetwork", "Force_Field", callback_force);
```
SyllogismRXS: I'm not sure the MotorSchemas behavior would have access to the external force measurement through a real sensor.

shaun-d-anderson: Eliminated change to keep MotorSchemas clean of project specific modifications.

src/plugins/controller/UnicyclePID/UnicyclePID.cpp
shaun-d-anderson: Done.

src/plugins/controller/UnicyclePID/UnicyclePID.cpp
shaun-d-anderson: Agreed.  Done.

PR #363, opened 2019-09-18
SimControl API change

SyllogismRXS:


include/scrimmage/simcontrol/SimControl.h
```diff
@@ -276,6 +462,33 @@ class SimControl {
 
     DelayedTask reseed_task_;
     bool limited_verbosity_;
+
+ private:
+    bool take_step();
```
SyllogismRXS: Moved many functions that were public to private to reduce complexity of SimControl API for external users.

include/scrimmage/simcontrol/SimUtils.h
```diff
@@ -92,16 +92,9 @@ boost::optional<std::string> run_test(const std::string &mission,
                                       const bool &init_python = true,
                                       const bool &finalize_python = true);
 
-bool logging_logic(MissionParsePtr mp, std::string s);
-
-std::shared_ptr<Log> preprocess_scrimmage(MissionParsePtr mp, SimControl &simcontrol);
```
SyllogismRXS: The logic in preprocess_scrimmage and postprocess_scrimmage were integrated into SimControl to reduce complexity of usage of SimControl

python/scrimmage/bindings/src/py_openai_env.cpp
```diff
@@ -248,51 +247,42 @@ void ScrimmageOpenAIEnv::scrimmage_memory_cleanup() {
 
 void ScrimmageOpenAIEnv::reset_scrimmage(bool enable_gui) {
     scrimmage_memory_cleanup();
-    mp_ = std::make_shared<sc::MissionParse>();
     simcontrol_ = std::make_shared<sc::SimControl>();
-    if (!mp_->parse(mission_file_)) {
+    if (not simcontrol_->init(mission_file_)) {
```
SyllogismRXS: Users of SimControl only have to create the SimControl instance, which creates its own instance of MissionParse and Log, so the user of SimControl doesn't have to create MissionParse and Log.

include/scrimmage/pubsub/NetworkDevice.h
```diff
@@ -73,19 +73,20 @@ class NetworkDevice {
     }
 
 
-    void set_max_queue_size(unsigned int size);
+    void set_max_queue_size(const unsigned int& size);
```
SyllogismRXS: Where applicable (and where it won't break user API), use of const-correctness. I didn't fix the entire scrimmage api, but I making progress as files are modified.

include/scrimmage/simcontrol/SimControl.h
```diff
@@ -64,64 +64,215 @@ enum class EndConditionFlags {TIME = 1, ONE_TEAM = 2, NONE = 3, ALL_DEAD = 4};
...
+    bool start();
+
+    /**
+     * @brief Runs the scrimmage simulation in the current thread.
+     *
+     * Example API usage:
+     *
+     *  \code{.cpp}
+     *  SimControl simcontrol;
+     *  simcontrol.init("./missions/straight.xml");
```
esquires: should `start` be mentioned in this comment? It seems like a tutorial is a better place for this documentation since users won't typically know to look here.

include/scrimmage/simcontrol/SimControl.h
```diff
@@ -64,64 +64,215 @@ enum class EndConditionFlags {TIME = 1, ONE_TEAM = 2, NONE = 3, ALL_DEAD = 4};
...
-    bool output_summary();
-    bool output_runtime();
-    void setup_timer(double rate, double time_warp);
-    void start_overall_timer();
-    void start_loop_timer();
-    void loop_wait();
-    void inc_warp();
-    void dec_warp();
-    void pause(bool pause);
+    /**
```
esquires: it seems like all of this comment and the others could just be a single line. doxygen formatting allows this with `///`

include/scrimmage/simcontrol/SimControl.h
esquires: good idea. Thanks.

include/scrimmage/pubsub/NetworkDevice.h
esquires: Thanks.

include/scrimmage/simcontrol/SimControl.h
SyllogismRXS: I'll add this to the tutorials. In this case, start() doesn't have be called because it is called automatically by run(). When using the run_single_step() function directly, the user has to call start() explicitly. My motivation for allowing these different processes is to make unit testing easier. Using the run_single_step() in a unit test allows the tester to check the state of the simulation at each time step without having to wait for the entire simulation to complete. This also allows the testing unit to inject published messages at specific simulation times.

include/scrimmage/simcontrol/SimControl.h
SyllogismRXS: Ok. Single line doxygen comments use /// now.

PR #366, opened 2019-09-20
add compiler flag to boost target

crichardson332:


crichardson332:
Before merging this in, we need to be sure this is necessary. Klimka had a build error from boost::ode that this fixed, but is this flag necessary? Normally I would just make this a PRIVATE compile option so it wouldn't affect anything else, but because boost is header only this flag will get set for almost everyone. What do you think?

PR #368, opened 2019-09-24
Added instructions for using AirSim's Asset environments and fixed typos.

nrakoski3:
Added instructions for using AirSim's Asset environments and fixed typos.


SyllogismRXS:
Thanks!

PR #377, opened 2019-10-10
Update jsbsim

crichardson332:
I'm going ahead and opening this pull request so you can take a look if you need. We should get this updated soon.


SyllogismRXS:
This was completed in a different pull request.

PR #397, opened 2019-11-20
remove broken class member calls in rosautonomy

dmagree:
this is a bandaid, open to suggestions for the proper fix


SyllogismRXS:
Fine for now. I have a task to improve our ROS integration anyway.

PR #427, opened 2020-03-13
Use Multiple Vehicles in AirSim

nrakoski3:
Can now have 1 vehicle in AirSim per Entity in Scrimmage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -274,7 +274,7 @@ set(ROS_VERSION "kinetic" CACHE STRING "The ROS distribution version.")
 if (BUILD_ROS_PLUGINS)
   set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH} /opt/ros/${ROS_VERSION}/)
   find_package(catkin REQUIRED COMPONENTS roscpp sensor_msgs std_msgs
-    geometry_msgs nav_msgs tf image_transport cv_bridge tf2_ros tf2)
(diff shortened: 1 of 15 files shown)


CMakeLists.txt
```diff
@@ -274,7 +274,7 @@ set(ROS_VERSION "kinetic" CACHE STRING "The ROS distribution version.")
 if (BUILD_ROS_PLUGINS)
   set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH} /opt/ros/${ROS_VERSION}/)
   find_package(catkin REQUIRED COMPONENTS roscpp sensor_msgs std_msgs
-    geometry_msgs nav_msgs tf image_transport cv_bridge tf2_ros tf2)
+    geometry_msgs nav_msgs tf image_transport cv_bridge tf2_ros tf2 tf2_geometry_msgs mavros_msgs)
```
SyllogismRXS: Can you move this reference to mavros_msgs to the ROSAirSim CMakeLists.txt? That way to build general ROS plugins, they don't need to install mavros_msgs as well. However, they will have to install mavros_msgs to use AirSim, I think this is fine.

nrakoski3: Hey Kevin I changed this, but if failed on the 16.04 build because "rosdep upate" couldn't run. Maybe the bionic key server was down? Could you try re-running the travis build on 16.04?

PR #428, opened 2020-03-19
IMU simulation

dburke3:
Port of an IMU simulator built in MatLab into the RosImuSensor to produce 'ideal' IMU data based on current and previous data frames of an entity. Modify csv data output for usage in other systems.

### DIFF
--- a/src/plugins/sensor/ROSIMUSensor/ROSIMUSensor.cpp
+++ b/src/plugins/sensor/ROSIMUSensor/ROSIMUSensor.cpp
@@ -50,6 +50,7 @@
 
 using std::cout;
 using std::endl;
+using namespace GeographicLib;
 
 namespace sc = scrimmage;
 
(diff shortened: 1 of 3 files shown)


nrakoski3:
Hey David please remove that extra set of sensor plugins in my quad-airsim-ex1.xml mission file under robot 2 so that scrimage gets that update in your pull request. Please fix your undefined symbol error so we know if we need to add a line to the ROSAltimeter's CMakeLists.txt file in this push. Please leave the Altimeter plugin in each of the robots in the mission file. Once you push those changes to your fork this pull request will be automatically updated. Thank you!

src/plugins/sensor/ROSIMUSensor/ROSIMUSensor.cpp
```diff
@@ -61,136 +62,408 @@ ROSIMUSensor::ROSIMUSensor() {}
...
+  double factor = sin(lat);
+  factor = sqrt(1 - ecc * ecc * factor * factor);
+  ecef(0) = (earth_radius / factor + alt) * cos(lat) * cos(lon);
+  ecef(1) = (earth_radius / factor + alt) * cos(lat) * sin(lon);
+  ecef(2) = ((earth_radius * (1 - ecc * ecc)) / factor + alt) * sin(lat);
+
+  return ecef;
+}
+
+Eigen::Vector3d ROSIMUSensor::ecef_to_lla(Eigen::Vector3d ecef) {
```
SyllogismRXS: We already have GeographicLib as a dependency for scrimmage. Can you take a look at the Forward() and Reverse() functions that it provides already to compute ECEF/LLA point conversions: https://geographiclib.sourceforge.io/html/classGeographicLib_1_1Geocentric.html

Unfortunately, GeographicLib doesn't provide orientation conversions, so you will have to perform those manually. By using libraries like GeographicLib we reduce the possibility of bugs creeping into the code base.

dburke3: Sounds good, I believe I incorporated GeographicLib where appropriate in my lastest commit, in addition to the forward and reverse conversions I also derived all of my constants from it.

PR #431, opened 2020-03-23
added a constant wind model for fixed wing and ardupilot interface

dmagree:
I'm getting some hits on the cpplint and cppcheck, but doesn't look like my stuff. Let me know if I need to make changes


SyllogismRXS:
Thanks!

PR #447, opened 2020-05-05
add operator* overload to quaternion

crichardson332:


src/math/Quaternion.cpp
```diff
@@ -154,4 +137,11 @@ std::ostream& operator<<(std::ostream& os, const Quaternion& q) {
     return os;
 }
 
+scrimmage::Quaternion operator*(const scrimmage::Quaternion &p, const scrimmage::Quaternion &q) {
+    Quaternion pq;
```
SyllogismRXS, line 141: I don't think pq is used. Right?

crichardson332: let's see if I can manage to stop screwing this up

SyllogismRXS:
Thanks for the update!

PR #452, opened 2020-05-18
Recompute simulated targets

dmagree:
In ContactBlobSensor, allow targets to be recomputed if the lat, lon, alt of the projection has changed.

### DIFF
--- a/src/plugins/sensor/ContactBlobCamera/ContactBlobCamera.cpp
+++ b/src/plugins/sensor/ContactBlobCamera/ContactBlobCamera.cpp
@@ -103,7 +103,6 @@ void ContactBlobCamera::parseSimulatedDetections(std::string &sim_det_str) {
             sim_contacts_[i++] = cnt;
         }
     }
-
 }
 
 void ContactBlobCamera::init(std::map<std::string, std::string> &params) {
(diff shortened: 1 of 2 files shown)


src/plugins/sensor/ContactBlobCamera/ContactBlobCamera.cpp
```diff
@@ -367,6 +375,14 @@ void ContactBlobCamera::add_false_positives(
 bool ContactBlobCamera::step() {
     if ((time_->t() - last_frame_t_) < 1.0 / fps_) return true;
 
+    if (proj_.LatitudeOrigin() != parent_->projection()->LatitudeOrigin() ||
```
SyllogismRXS, line 377: I'm not sure this is a scalable solution since it sort of breaks the idea that a scrimmage simulation takes place in a single ENU frame. I'm fine with translations to external systems that use lat/lon/alt, but a scrimmage plugin shouldn't be changing frames. Is there a way that you can manage this externally?

dmagree: If it makes a difference to you, this plugin does not ever change the projection, it only monitors for a change and recomputes targets when it does.

dmagree:
Found a different way to accomplish this

PR #461, opened 2020-06-19
Update the tutorial with GeoJSON content

olasanni1:
Add additional information to the terrain tutorial about creating
a GeoJSON file that contains ground-based polygon extrusion data.

Add troubleshooting information for the GeoJSON feature


SyllogismRXS:
Thanks!

PR #464, opened 2020-06-24
Team cpa plugin

tm132:
Created TeamCPA plugin (metrics) from CPA (metrics) plugins.
- computes CPA within a specific team (friendly CPA)
- allows specification of `team_id` in mission file (default 1)
- allows specification of `after_time_s` in the mission file (default 0.0), which directs the CPA to only be computed after the specified simulation time
- if `team_id` or `after_time_s` doesn't exist, nothing will be printed


tm132:
JD integrated these functionalities directly into the CPA plugin itself.

PR #468, opened 2020-07-15
add mission python module to package

crichardson332:


SyllogismRXS:
I actually have this fixed in another branch using the python find_packages() command, which is a more scalable solution to  manually listing the packages. I'm going to merge that in today.

PR #477, opened 2020-08-03
Potential Fix for seg-fault on GUI closing

jlewis9000:
This fix addresses a segfault I encountered trying to run scrimmage on windows using VTK 9.0.0. This issue doesn't seem to occur with older versions of VTK on Ubuntu. 

The seg-fault occurs when the window is closed. Debugging with gdb showed that the seg-fault occurs when the vtkSmartPointer destructor is called. The seg-fault doesn't seem to occur if Finalize is called before the destructor.


SyllogismRXS:
Thanks for the input, I included your changes the following pull request, which has been merged into the master branch: https://github.com/gtri/scrimmage/pull/478

PR #499, opened 2020-11-24
Cesium web server

crichardson332:


SyllogismRXS:
A nice addition, but we should limit the binary data files that are being added the core scrimmage repo. A better solution would be to add the binary files to the scrimmage-extra-data repo: https://github.com/SyllogismRXS/scrimmage-extra-data

crichardson332:
Yeah we still need to figure out the best way to allow plugin projects to provide 3d models and define their own custom mappings from entities to 3d models. Right now the 3d models have to be served with the static webpage in order to be accessed in cesium, and I don't know how to get extra 3d model files in the web page that aren't located in the same web src folder as the rest of the website.

crichardson332:
The cesium people will host up to 5G of tile data for you with a free Cesium Ion account, so if we had a scrimmage or GTRI account we could have a set of common models hosted there and not have to worry about this. Someone would just have to manage that account. And of course, if you pay for a commercial account then you get a lot more.

frazierbaker:
I have some concerns about seasocks given that it is linux-only.  Are we comfortable with making this functionality only available on linux?

PR #502, opened 2020-12-28
Bugfix/build script updates

christomaszewski:
Just updated the installation procedures a bit to match the current arguments to the install-binaries.sh script. Also changed the way the arguments are handled by that script to handle any ordering or weird combinations users might come up with.

Replying to comments on previous PR:
- The updated script is written such that you can pass in extra arguments like -p and -e without breaking anything. As the script currently is, I don't think those flags actually do anything. I can add a check to look for those flags as well in addition to --external or --python respectively, but that would actually be introducing a change. (I stumbled on this because I was trying to install python 3 support and -p 3 was not doing anything, it was still defaulting to installing both 2 and 3 support).

- The version checking line ( ! echo "$UBUNTU_VERSION 18.04 -p" | dc | grep > \dev\null ^-) is future compatibility. It should trigger if the Ubuntu version is greater than or equal to 18.04. It definitely could be more readable so if anyone has any other suggestions for how to do this check more cleanly just let me know and I'll make the change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,12 +93,13 @@ GenerateSetEnv(
 option(EXTERNAL "EXTERNAL" 0)
 
 option(ENABLE_PYTHON_BINDINGS "ENABLE_PYTHON_BINDINGS" ON)
-set(PYTHON_MIN_VERSION "2.7" CACHE STRING "Minimum Python version for building")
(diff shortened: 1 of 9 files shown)


christomaszewski:
Changes to tutorials got pulled into this. I'll fix that. Will submit tutorial updates in separate PR

PR #504, opened 2020-12-30
Fix to SimControl::finalize function to correct SimpleCollisionMetrics output

christomaszewski:
Added a final call to run_networks() and then trigger the metric callbacks after the EntityPresentAtEnd messages are published in SimControl::finalize(). This fixes the bug where these messages never arrive to the SimpleCollisionMetrics plugin, which causes all teams to always show up as not having survived. 

This fix also seems to correct the total flight time metric for the single agent during runs of the straight mission when it survives. The length of the straight mission is 100 seconds, so if the single agent team survives, the expected total flight time for that team is also 100. Prior to making this change I was getting total flight times of 53 (I guess this might vary but it seems pretty stable for me) when the single agent team survives. After the fix the total flight time shows 99.9.

### DIFF
--- a/src/simcontrol/SimControl.cpp
+++ b/src/simcontrol/SimControl.cpp
@@ -1016,6 +1016,11 @@ bool SimControl::finalize() {
         pub_ent_pres_end_->publish(msg);
     }
 
+    // Run the networks one final time to send out the messages sent above
+    // Then trigger the metric callbacks to run 
+    run_networks();
+    br::for_each(metrics_, run_callbacks);
+
     run_logging();
 
     if (display_progress_) cout << endl;


christomaszewski:
Any thoughts on this fix? I'd like to update the tutorials so that new users can follow along and match their output with the output shown in the tutorials when running with a deterministic seed. Are metrics plugins just generally not used or has someone fixed this issue on a private repo already? It seems like this might cause issues if you rely on all the messages sent in finalize actually getting through to the metrics plugin.

christomaszewski:
Can someone trigger CI on this again. There's no reason this should be failing on Ubuntu 20.04.

mday299:
Did this PR fix it?
https://github.com/gtri/scrimmage/pull/511

christomaszewski:
> 
> 
> Did this PR fix it?
> #511

I can give it a shot but I don't see anything in the commits from that PR that specifically addresses the issue. I'll try to test it and report back.

tm132:
> 
> 
> Can someone trigger CI on this again. There's no reason this should be failing on Ubuntu 20.04.

The CI on 20.04 failed due to docker rate limit, which isn't really related to the code changes. Admins (@mday299 or @frazierbaker etc), has the CI since this last failed been updated to work around this problem? If so, can one of the admins please kick off a rebuild to get this passing? This seems like a change that should be merged in.
https://travis-ci.com/github/gtri/scrimmage/jobs/469189352

```
library/ubuntu

toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

The command "docker build -t $REPO:focal-${TRAVIS_BUILD_NUMBER} -f ./ci/dockerfiles/ubuntu-20.04 ." exited with 1.

0.00s$ export TAG=`if [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo "latest"; else echo $TRAVIS_BRANCH ; fi`
```

wfsyre:
Fixed by #605

PR #507, opened 2021-01-26
fixed nullptr in draw_sphere

wfsyre:
FIxed an issue where SafeDowncast produced a nullptr and caused a segfault.

### DIFF
--- a/src/viewer/Updater.cpp
+++ b/src/viewer/Updater.cpp
@@ -2264,6 +2264,9 @@ bool Updater::draw_sphere(const bool &new_shape,
     } else {
         sphereSource = vtkSphereSource::SafeDownCast(source);
     }
+    if (!sphereSource) {
+        return false;
+    }
     sphereSource->SetRadius(s.radius());
     actor->SetPosition(s.center().x(), s.center().y(), s.center().z());
     return true;


SyllogismRXS:
In what context was the nullptr encountered? I am inclined to not have this nullptr check here because the object should be safe by design at this point. Since that wasn't true in your case, it may point to a bigger problem that needs to be addressed. Many of the other "draw_" functions use the same control structure for down casting the shape or creating the shape if it's new. Could it have been your case that the shape wasn't "new," but it was not originally a sphere shape? If so, we may need to handle that case better at a higher level to keep this error from popping up when using other shapes.

wfsyre:
This bug only happens when I draw > 40 spheres, and occurs right after the call to draw_shape(), so I'm pretty sure a new sphere is somehow being picked up as an old shape. The crash happens consistently in the mission I'm running, but not at a consistent time step, or caused by any discernible event. My original thought was that there was a hash collision with an old shape. I agree that this error could likely be picked up earlier in the process, but I'm unsure where to start looking. For starters, I've checked the shapes that make it into the EntityPlugin class and they seem correctly formatted.

wfsyre:
I finally got some time to dig into this. The bug is the result of a hash collision, detected in Updater.cpp. This is only possible if the two offending shapes are drawn on the same timestep, by the same parent Entity, and get the exact same random number from rng_uniform. In my opinion, this isn't really solvable in Updater.cpp because setting a new hash there would be difficult since we don't have access to the original shape. We'd have to keep some state about a remapped hash and the shape we are remapping. So instead I'm solving it in EntityPlugin by multiplying 2 rng_uniforms by each other to increase the randomness of the hash. Hopefully no two shapes get the same two random numbers.

PR #514, opened 2021-02-23
added more randomness to the entityPlugin shape hash

wfsyre:


wfsyre:
This will at least make the hash collision problem I experienced more infrequent until/if larger changes are deemed necessary.

mday299:
Merged! Thanks.

Issue #440, opened 2020-04-03
env default python should be standard, not 2.7


tm132:
Scrimmage's default cmake configuration should use the default version of python setup in the user's environment (hopefully 3.x), instead of quietly using python 2.7, since 2.7 is deprecated.

i.e., if you run...
`$ cmake ..`
...it should build with whatever version of python they have selected as the default in their environment.

For instance, mine is set to:
```
$ python -V
Python 3.7.3
$ which python
/home/user/anaconda3/bin/python
$
```

-+-

The cmake files should be updated, and then the README.md should be updated, something like this, if you wish to allow users to target other/older versions of python:

> You can specify a target Python version in SCRIMMAGE core by setting the PYTHON_TARGET_VERSION cmake variable. For example, to specify a Python version of 2.7, first clear the cmake cache file and rerun cmake:
```
$ rm CMakeCache.txt
$ cmake .. -DPYTHON_TARGET_VERSION=2.7
```

mday299:
@tm132 finally getting to this.  Is this still an issue?

tm132:
No one contacted me saying they were fixing it, so it probably still is. Glancing over the CMakeLists.txt file, it looks like it might still be a problem. I might have time to test this in the next couple of days to re-verify.

tm132:
An update: this problem still exists:

```
scrimmage@orca:~/sandbox/scrimmage/build$ cmake ..
-- The C compiler identification is GNU 7.5.0
-- The CXX compiler identification is GNU 7.5.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
// HERE -----------------------------------------------
-- Found PythonInterp: /usr/bin/python2.7 (found suitable version "2.7.17", minimum required is "2.7")
-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython2.7.so (found suitable version "2.7.17", minimum required is "2.7")
-- Python Versions Found: 2.7.17
[...]

PR #521, opened 2021-03-30
CMake updates to use environment Python by default

tm132:
Closes: #440 
Changes:
- Primarily, the main `CMakeLists.txt` was updated to remove the setting of `PYTHON_MIN_VERSION` to 2.7 by default.
- The `README.md` was updated to clarify what the `PYTHON_MIN_VERSION` flag is used for.

Functionality changes:
- `PYTHON_MIN_VERSION` was repurposed to not be set to anything by default, and instead allows the user to optionally specify, if they would prefer, a different python version to use than the environment default. For example, the latest anaconda (2020.11) comes with python 3.8.
- By default (no variables specified, e.g., `cmake ..`), CMake will use the environment default python version.
- Any previous scripts using this flag to automate the building of scrimmage should be backwards-compatible with these changes.

Testing:
```
$ cd build/
$ rm CMakeCache.txt
$ cmake ..                             # should use env default Python
$ rm CMakeCache.txt
$ cmake .. -DPYTHON_MIN_VERSION=2.7    # should use Python 2.7
$ rm CMakeCache.txt
$ cmake .. -DPYTHON_MIN_VERSION=3.4    # should use Python 3.7 (interpreter), 3.6 (libs) on Ubuntu 18.04 vanilla
$ rm CMakeCache.txt
$ cmake .. -DPYTHON_MIN_VERSION=3.8    # on systems with anaconda 2020.11 installed and properly configured in their path, should use Python 3.8
```

Implemented and tested on a vanilla Ubuntu 18.04 VBox VM.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,24 +93,16 @@ GenerateSetEnv(
 option(EXTERNAL "EXTERNAL" 0)
 
 
-set(PYTHON_MIN_VERSION "" CACHE STRING "Deprecated; please use PYTHON_TARGET_VERSION instead.")
-if (PYTHON_MIN_VERSION)
-  message(DEPRECATION "!!!!!  Variable 'PYTHON_MIN_VERSION' is deprecated.\
-                       The environment default python will be used. If you\
-                       wish to specify a specific python version, please\
-                       use 'PYTHON_TARGET_VERSION' instead. !!!!!")
-endif()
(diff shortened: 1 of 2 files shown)


mday299:
@tm132 I'm hesitant to include this change at this stage. A lot of Linux distros are hanging on to Python 2 still.

PR #522, opened 2021-03-31
allowing an external force in the fixed wing model

dmagree:
allows external force to be applied and fixes the transport function


frazierbaker (approved):
Nihil obstat from my perspective.  I want to see what the CI thinks before merging to make sure.

dmagree:
Any word on the CI results?

PR #523, opened 2021-04-03
WIP: Migrate to GitHub CI

frazierbaker:


frazierbaker:
The GitHub CI should work for free with a couple of caveats:
1. The build time cannot exceed 6 hours.
2. The docker pushes will not work without credentials to upload to dockerhub.  I have a multiplatformautonomy organization set up on DockerHub that is slightly more generic, though it may be worth having GTRI setting up a GTRI dockerhub organization and supplying build keys via a GitHub secret (GitHub secrets will be redacted in any logs they show up in).  If GTRI doesn't want to do this, then I need admin access to the repository to add my generated DockerHub credentials as a secret.

frazierbaker:
It looks like the test that fails is actually in https://github.com/SyllogismRXS/scrimmage_ros.git

https://github.com/SyllogismRXS/scrimmage_ros/blob/master/test/test_auction.cpp

SyllogismRXS:
Maybe ask @esquires about the test_auction fail. Pretty sure he put that one together.

SyllogismRXS:
@frazierbaker Is GTRI going to setup a DockerHub account? I think that would be ideal.

frazierbaker:
@SyllogismRXS I agree that that would be ideal, we are working on that.  @esquires any advice on the rostests?

esquires:
I am not sure unfortunately. I would run the test individually to see the cout.

wfsyre:
CI working on 24.04 branch

Issue #1, opened 2017-08-04
add shapes to plugin base


esquires:
Right now autonomies and entity interactions can draw shapes. It would be nice if sensors/network/etc plugins could do so as well.

SyllogismRXS:
all plugins can draw shapes now, but the sensor interface needs to be updated to be callable by simcontrol. This will be resolved by issue #74

PR #528, opened 2021-06-10
Fix bug #1

kemen209:
Use stoi to get int type of motion_multiplier_


frazierbaker:
What was the issue this was intended to fix?

kemen209:
To avoid implicitly type conversion from double to int.

tm132:
`motion_multiplier_` is defined as an `int` in the header file, so stoi would be correct in this context.

However, we have this in the same file:
```
motion_multiplier_ = 1;    // TM32: an int?
[......]
double MissionParse::motion_multiplier() { return motion_multiplier_; }   // TM132: a double??
```

Should it be a `double` or should it be an `int`?

wfsyre:
Fixed in another commit

Issue #557, opened 2022-09-30
Update plot_3d_fr.py


tm132:
Updates to `plot_3d_fr.py` (quality of life updates):
- ~~fix bug that prevents it from working~~ removed this, as there are bigger issues here that need to be looked at,
- add a few new options
  - `-i` for interactive mode (this needs to be runnable without the gui so that it can be used as part of an automated process)
  - -`o <output directory>` (allow the dir where the output plots are written to to be specified)
  - Change the default output dir to the location where frames.bin is found

PR #558, opened 2022-10-01
Issue 557 updates to plot 3d fr py

tm132:
Closes: https://github.com/gtri/scrimmage/issues/557

Updated `scripts/plot_3d_fr.py` - Quality of Life updates:
- ~~fixed an index error bug~~ removed this as there are bigger issues at play that need to be looked into more in-depth
- added `-i` option for interactive mode. The gui will no longer be displayed by default. If you want to run the gui, use the `-i` (interactive) flag from now on.
- made the default output directory the same directory that `frames.bin` was found in.
- added `-o` (output directory) option which will allow the user to specify an alternative output directory for writing out the plots.

Updated proto_utils.py
- make the error message in read_frames() more detailed/descriptive

These updates were made for the purpose of allowing this plotting script to be used in an automated process (switching it from interactive-by-default to automatic-by-default (no gui) as well as adding an optional command line option to specify the output directory of your choice).

### DIFF
--- a/python/scrimmage/proto_utils.py
+++ b/python/scrimmage/proto_utils.py
@@ -68,8 +68,8 @@ def read_frames(frames_file, to_dataframe=False):
 
         try:
             frame.ParseFromString(data[pos:pos + next_pos])
-        except:
-            print('Error reading frames! Aborting data read-in.')
+        except Exception as e:
+            print(f'Error reading frames!\nError: {e}\nAborting data read-in.\n')
             break
 
         frames.append(frame)
--- a/scripts/plot_3d_fr.py
+++ b/scripts/plot_3d_fr.py
@@ -160,7 +160,7 @@ def main():
 
     for fr in frames: # at each timestep
         for con in fr.contact: # for each contact/entity
-            conidx = min(con.id.id - 1, 0)
+            conidx = con.id.id - 1
             # append each entity's x, y, z position to the appropriate list
             allentX[conidx].append(con.state.position.x)
             allentY[conidx].append(con.state.position.y)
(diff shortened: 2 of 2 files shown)


tm132:
While I think there are some good QoL updates here (the updates to the invocation options), there is something fundamentally wrong with how this seems to attempt to plot data that likely has at least something to do with the `index out of bounds error` line. It is using `contact.id.id-1` instead of just `contact.id.id`, making the assumption that `entity_id` will never be lower than `1`, and then shifting all entity ids down by 1 and using them as the index of an array. Some simulation outputs I was testing with were causing an `index out of bounds` error because of this assumption.

The fix for this `index out of bounds` error in the PR is a band-aid, not a real robust solution. It should probably be replaced with a map(), or have a more in-depth reimplementation, or maybe even just re-evaluated for overall usefulness.

Because this was producing plots that weren't right, instead of fixing this utility more completely, I instead used it as an example of how to interact with scrimmage's python package for parsing frames.bin, and incorporated a new frames.bin translator into my own project's post processor tool, which is producing correct-looking plots from the frames.bin data.

I'm on the fence about what to do about this one: 
- revert the fix `conidx = min(con.id.id - 1, 0)` back to the original `conidx = con.id.id - 1` and then merge in the PR to get all the QoL updates
- just decline this PR and move on

tm132:
I decided to go ahead and remove the one-line fix for the `index out of bounds` error I was getting. The only thing remaining in the PR are the QoL updates. This PR can be merged if the QoL updates are desireable.

wfsyre (approved):

PR #569, opened 2023-04-24
Adding Play and Pause Indicator on the Scrimmage Simulation

laserjetprinter:
WIP


laserjetprinter:
Move the play/pause to the bottom right corner

PR #568, opened 2023-04-10
Add Plugin XML Attributes to Mission XML Log

laserjetprinter:
WIP

PR #572, opened 2023-05-19
Mission to Mission with Plugin XML to Mission XML

laserjetprinter:
Combining these pull requests, because the mission to mission pull request will require code from the current open pull request below:

https://github.com/gtri/scrimmage/pull/568/files


laserjetprinter:
This didn't combine like expected... closing for now

PR #574, opened 2023-07-11
check for ros master in a separate thread

dmagree:
This change periodically checks for availability of ros master, and reconnects topics if necessary. This allows rosmaster to be restarted without affecting scrimmage simulation.


src/plugins/interaction/ROSClockServer/ROSClockServer.cpp
```diff
@@ -73,16 +73,40 @@ bool ROSClockServer::init(std::map<std::string, std::string> &mission_params,
...
-    // Tell other nodes to use the simulated time from "/clock"
-    ros::param::set("/use_sim_time", true);
+    return true;
+}
 
-    // Publish the first time message
-    publish_clock_msg(time_->t());
+void ROSClockServer::close(double t) {
+    (void) t;
+    std::cout << "ROSClockServer: closing" << std::endl;
```
frazierbaker, line 85: Is this for debugging or is it actually necessary in production?

src/plugins/interaction/ROSClockServer/ROSClockServer.cpp
```diff
@@ -73,16 +73,40 @@ bool ROSClockServer::init(std::map<std::string, std::string> &mission_params,
...
+        if (prev_rosmaster_state_ == false && rosmaster_state == true) {
+            // time to reinitialize
+            nh_ = std::make_shared<ros::NodeHandle>(); // this actually starts the node
+            // Create Publisher
+            pub_mutex_.lock();
+            clock_pub_.shutdown();
+            clock_pub_ = nh_->advertise<rosgraph_msgs::Clock>("/clock", 1);
+            ros::param::set("/use_sim_time", true);
+            pub_mutex_.unlock();
+            std::cout << "ROSClockServer: Ros initialized!" << std::endl;
```
frazierbaker, line 102: Are these `cout` statements necessary, or were they for debugging?  If only for debugging, please remove.

src/plugins/interaction/ROSClockServer/ROSClockServer.cpp
```diff
@@ -73,16 +73,40 @@ bool ROSClockServer::init(std::map<std::string, std::string> &mission_params,
         ros::init(argc, NULL, "scrimmage", ros::init_options::NoSigintHandler);
     }
 
-    nh_ = std::make_shared<ros::NodeHandle>();
-    clock_pub_ = nh_->advertise<rosgraph_msgs::Clock>("/clock", 1);
+    // rosmaster check thread
+    check_rosmaster_thread_ = std::thread(&ROSClockServer::check_rosmaster, this);
+    running_ = true;
```
frazierbaker, line 78: `std::thread` begins execution immediately upon construction, but running_ is set to true after construction.  This could result in a race condition where `while(running_)` evaluates to false and nothing happens in the thread before it returns.  Please consider moving `running_ = true;` to the line before the thread initialization.

frazierbaker (changes requested):
I think I identified a race condition here that needs to be addressed.  There are some other questions on `std::cout` statements that seem to be leftover from debugging.

dmagree:
I've removed the printouts and fixed the race condition. I've also added one synchronous call in init to try to connect to rosmaster in order to preserve the previous behavior if there is already a ros master running

frazierbaker:
I'm confused by the ROS Master Check and how it preserves existing behavior.  What would I have to do differently to get the new behavior?  If I run a relevant mission file written before this pull request, will I get the new behavior or the previous behavior?

dmagree:
You get both! The difference from what I had previously it that now you are ensured to get one attempt to connect to ros master during init. Before, the thread was started in the init, but there was no explicit guarantee that the function in the thread would finish before the first call to update, so you could get a full time step not connected to master, at least in principle. Now, you have one call to connect to master that in synchronous, just like prior to this merge, and only after this do you start the thread to continuously check whether you need to (re)connect to master.

PR #587, opened 2024-07-12
FixedWing6DOF: more robust rotation handling

dmagree:
This request fixes numerical precision issues with converting quaternions to roll pitch yaw in the model.


wfsyre (changes requested):
Moving the logic from this one particular implementation to the Quaternion pitch() method because it seems to be a larger issue.

dmagree:
Instead of modifying the `pitch()` call, I've created a `pitch_safe()` member function that restricts the argument of the `asin()` call to be in the range [-1, 1].

wfsyre (approved):

PR #600, opened 2024-12-04
enable external playback

esquires:


wfsyre:
Is the purpose of this to allow someone to press space and begin taking screenshots every step? Can we bind the action  to something other than the space key?

esquires:
Good question. The main purpose is to make it easier to use external simulation tools for dynamics but scrimmage as a gui. The space key was there before to step the simulation and was used to enable taking screenshots. The current behavior is that when screenshots are setup in the mission file, pressing space at the beginning of a mission will start a loop where scrimmage proceeds by taking a screenshot at every timestep.

Part of allowing an external tool to do the simulation but not the gui was making it easier to use the scrimmage playback tool to take screenshots. The only change to main.cpp is to leverage the above behavior by setting it to proceed a step at a time to take screenshots, similar to how the scrimmage binary works.

The other change to make it easier to use an external tool is in `Log.h` by allowing external tools to write individual proto files. The packing algorithm in scrimmage into a single file wasn't easy to recreate in other languages so it seems easier to allow users to write protobuf files in a more natural way.

We can certainly use a different key if desired but for consistency should probably also update normal scrimmage runs to use a different key as well. I think it is probably best long term to get "b" and space interacting with screen saving a bit easier to use. This pull request just keeps the behavior consistent in the playback tool though.

wfsyre (approved):

PR #625, opened 2026-01-02
change csv to deal in strings now and convert appended values into strings

wfsyre:
Right now, the CSV writer class can only handle writing doubles to the csv table object. These get converted into strings when writing to the actual CSV. To simplify things, the CSV writer should just maintain a map of strings, which allows us to write any data type we want to the CSV as long as we convert it to a string first. It introduces some extra pre-processing we have to do to recover data from a CSV, but I think the schema of `get<datatype>()` is already well established in scrimmage so its not a large burden. This also allows people the added control of determining how their data should be turned into a string instead of leaving it to the CSV writer to determine. I have left the scientific vs fixed schema in there but technically all that should be done before the CSV writer even sees the value. The Visitor stuff is a bit of a hacky way to leave some backward compatibility for the old Append method that only took a double.

### DIFF
--- a/include/scrimmage/common/CSV.h
+++ b/include/scrimmage/common/CSV.h
@@ -53,12 +53,10 @@ struct StringifyVisitor {
     bool double_is_scientific = true;
     int double_precision = 13;
 
-    std::string operator()(int value) const { return std::to_string(value); }
-    std::string operator()(unsigned int value) const { return std::to_string(value); }
-    std::string operator()(size_t value) const { return std::to_string(value); }
-    std::string operator()(long value) const { return std::to_string(value); }
+    std::string operator()(bool value) const { return value ? "true" : "false"; }
+    std::string operator()(uint64_t value) const { return std::to_string(value); }
+    std::string operator()(int64_t value) const { return std::to_string(value); }
     std::string operator()(const std::string& value) const { return value; }
-    std::string operator()(bool value) const { return value ? "True" : "False"; }
     std::string operator()(double value) const {
         // default precision values for double are not enough in many cases
         std::ostringstream conv;
@@ -79,8 +77,7 @@ namespace scrimmage {
 class CSV {
  public:
     typedef std::list<std::string> Headers;
-    typedef std::variant<bool, size_t, unsigned int, long, int, double, std::string>
-        PossibleVariantTypes;
+    typedef std::variant<bool, uint64_t, int64_t, std::string, double> PossibleVariantTypes;
     typedef std::list<std::pair<std::string, PossibleVariantTypes>> Pairs;
 
     ~CSV();
--- a/include/scrimmage/common/CSV.h
+++ b/include/scrimmage/common/CSV.h
@@ -34,18 +34,54 @@
 #define INCLUDE_SCRIMMAGE_COMMON_CSV_H_
 
 #include <fstream>
+#include <iomanip>
+#include <iostream>
 #include <list>
 #include <map>
 #include <memory>
+#include <sstream>
 #include <string>
 #include <utility>
+#include <variant>
+
+#include "scrimmage/parse/ParseUtils.h"
+
+namespace {
+struct StringifyVisitor {
+
+    bool double_is_fixed = true;
+    bool double_is_scientific = true;
+    int double_precision = 13;
+
+    std::string operator()(int value) const { return std::to_string(value); }
+    std::string operator()(unsigned int value) const { return std::to_string(value); }
+    std::string operator()(size_t value) const { return std::to_string(value); }
+    std::string operator()(long value) const { return std::to_string(value); }
+    std::string operator()(const std::string& value) const { return value; }
+    std::string operator()(bool value) const { return value ? "True" : "False"; }
+    std::string operator()(double value) const {
+        // default precision values for double are not enough in many cases
+        std::ostringstream conv;
+        if (double_is_fixed) {
+            conv << std::fixed;
+        }
+        if (double_is_scientific) {
+            conv << std::scientific;
+        }
+        conv << std::setprecision(double_precision) << value;
+        return conv.str();
+    }
+};
+}  // namespace
 
 namespace scrimmage {
 
 class CSV {
  public:
     typedef std::list<std::string> Headers;
-    typedef std::list<std::pair<std::string, double>> Pairs;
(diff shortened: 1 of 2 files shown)


include/scrimmage/common/CSV.h
```diff
@@ -34,18 +34,54 @@
...
+    bool double_is_fixed = true;
+    bool double_is_scientific = true;
+    int double_precision = 13;
+
+    std::string operator()(int value) const { return std::to_string(value); }
+    std::string operator()(unsigned int value) const { return std::to_string(value); }
+    std::string operator()(size_t value) const { return std::to_string(value); }
+    std::string operator()(long value) const { return std::to_string(value); }
+    std::string operator()(const std::string& value) const { return value; }
+    std::string operator()(bool value) const { return value ? "True" : "False"; }
```
jblee123: I'd opt for "true" and "false" (not capitalized). There's not a *huge* reason for this, but there's also not a huge reason for capitalizing, so my "break the tie" reasoning" is that if people are ever typing it, that's one less thing to mess up. If it's read case-insensitive (which I would hope it is), then no reason to throw in a capital and make people wonder. Python is the only thing I know of that has case-sensitive "True" and "False". No reason to bring that into a CSV.

include/scrimmage/common/CSV.h
```diff
@@ -34,18 +34,54 @@
...
+};
+}  // namespace
 
 namespace scrimmage {
 
 class CSV {
  public:
     typedef std::list<std::string> Headers;
-    typedef std::list<std::pair<std::string, double>> Pairs;
+    typedef std::variant<bool, size_t, unsigned int, long, int, double, std::string>
```
jblee123: Given possible ripple effects, I'm not sure if what I'm about to say is feasible. Actually, if all you had was `double` before, then all these options should be new, so it wouldn't affect existing code any differently, so I'll go with that assumption. Try to erase unsized types like `unsigned int`, `long`, `int`, etc. from usage. Prefer sized types. What's an `int`? It has a minimum size, but you don't really *know* because those are implementation-defined. Same for `long`. I'm guessing for `int` you're expecting a 32-bit value, so use `int32_t`. I'm guessing `long` is expected to be a 64-bit value, so use `int64_t`.

And if these are for data files, you can probably get away without a `size_t`. Depends on what you're expecting to represent. A `size_t` isn't "how many things", but is specifically a value used to represent memory sizes, so it's tied to the size of a pointer. That's why they're used as offsets in arrays -- not because they're counting items, but because they're being used as memory offsets. So while you use it in code, I'm not sure if you'd want that kind of semantic exposed as an option in a data file.

include/scrimmage/common/CSV.h
```diff
@@ -75,7 +111,11 @@ class CSV {
 
     size_t rows();
 
-    double at(int row, const std::string& header);
+    template <class T1>
+    T1 at(int row, const std::string& header) {
```
jblee123, line 112: see -- now *this* is where I'd use a size_t, because it's an index. But this ties into existing underlying definitions, so I wouldn't change that now. (Actually, taking a closer look... the table uses an int as a key into a map... wut? Why is it not a vector? And can you have negative column/row indexes? Weird... still -- not something for *this* update.)

src/common/CSV.cpp
```diff
@@ -73,14 +69,23 @@ void CSV::set_column_headers(const std::string& headers, bool write) {
...
+            .double_is_fixed = double_is_fixed_,
+            .double_is_scientific = double_is_scientific_,
+            .double_precision = double_precision_},
+        v);
+}
+
 bool CSV::append(const Pairs& pairs, bool write, bool keep_in_memory) {
 
-    for (std::pair<std::string, double> pair : pairs) {
+    for (auto pair : pairs) {
```
jblee123: The existing code was making a copy of the pair every time, and the current code still does that. You want to iterate over the pairs by reference, so `auto&`, and if you're not changing anything, which you're not since `pairs` is `const`, then `const auto&`.

src/common/CSV.cpp
```diff
@@ -163,22 +168,7 @@ std::string CSV::row_to_string(const int& row) const {
...
-            }
-            if (double_is_scientific_) {
-                conv << std::scientific;
-            }
-            conv << std::setprecision(double_precision_) << kv.second;
-            values[kv.first] = conv.str();
-        } else {
-            values[kv.first] = std::to_string(kv.second);
-        }
+        values[kv.first] = kv.second;
```
jblee123: This will work, but check the fuller context. Originally, the logic was: 1) build a vector of converted strings, 2) build an aggregate string out of those. You already have the strings, though, so your `values` vector here is really just making a bunch of copies. You could get rid of the copies by making it a vector of string ptrs (or references, but it'd have to be of `reference_wrappers` because you can't make a vector of raw references, but ptrs would do fine in this case). But you can get rid of `values` altogether; it was never really needed in the first place since you can append to the final string as you iterate. No real need for an intermediate list. And continually appending to a string like this isn't the most efficient since it's having to re-allocate a bigger and bigger string every time. Make `result` a `std::ostringstream` of appending is more efficient to the stream, and then return `result.str()` so it's converted only once at the end.

src/common/CSV.cpp
```diff
@@ -238,7 +228,7 @@ bool CSV::read_csv_from_string(const std::string& csv_str, const bool& contains_
             std::vector<std::string> tokens;
             boost::split(tokens, line, boost::is_any_of(","));
             for (unsigned int i = 0; i < tokens.size(); i++) {
-                table_[row_num][i] = std::stod(tokens[i]);
+                table_[row_num][i] = tokens[i];
```
jblee123, line 228: This is going to be the most annoying thing, and is going to tie into `CSV::row_to_string()`... what happens if somebody writes a string that contains a comma? Look at the specification section at https://en.wikipedia.org/wiki/Comma-separated_values, and it tells you how to handle that situation. Any time you decide to allow strings in data files, that opens up "fun" edge cases that you have to think about. What about newlines? Carriage returns? Maybe `CSV::append()` needs to have more failure cases.

Note that when you handle things like escaped double quotes, those need to be handled via linearly scanning through the string; you can't do global substring substitutions because that may not work the way you expect it to for some cases.

wfsyre:
Closed by accident. Reopened as a new MR here #630

PR #630, opened 2026-03-30
Change csv to deal in strings and convert appended values into strings

wfsyre:
Right now, the CSV writer class can only handle writing doubles to the csv table object. These get converted into strings when writing to the actual CSV. To simplify things, the CSV writer should just maintain a map of strings, which allows us to write any data type we want to the CSV as long as we convert it to a string first. It introduces some extra pre-processing we have to do to recover data from a CSV, but I think the schema of get<datatype>() is already well established in scrimmage so its not a large burden. This also allows people the added control of determining how their data should be turned into a string instead of leaving it to the CSV writer to determine. I have left the scientific vs fixed schema in there but technically all that should be done before the CSV writer even sees the value. The Visitor stuff is a bit of a hacky way to leave some backward compatibility for the old Append method that only took a double.


src/common/CSV.cpp
```diff
@@ -44,6 +42,121 @@
 
 namespace scrimmage {
 
+// Function to escape special characters in a CSV field
+std::string escape_csv_field(const std::string& field) {
+    std::string escapedField = field;
+
+    // If the field contains a comma, a quote, or a newline, wrap it in quotes
+    if (field.find(',') != std::string::npos || field.find('"') != std::string::npos
```
jblee123: Minor, but formatting can help with readability. I now it's one more line, but
```
    if (field.find(',') != std::string::npos
        || field.find('"') != std::string::npos
        || field.find('\n') != std::string::npos) {
```
reads more easily than
```
    if (field.find(',') != std::string::npos || field.find('"') != std::string::npos
        || field.find('\n') != std::string::npos) {
```
Having said that, though, you can search for multiple chars at once with:
```
    if (field.find_first_of(",\"\n") != std::string::npos) {
```

src/common/CSV.cpp
```diff
@@ -44,6 +42,121 @@
...
+        || field.find('\n') != std::string::npos) {
```
jblee123: If you find the entire meat of a function is inside an `if` statement, that's a clue that an early-out model would have been good, so:
```
    bool need_to_escape = field.find_first_of(",\"\n") != std::string::npos;
    if (!need_to_escape) {
        return field;
    }

    // do the escape logic
    return escaped.str();   
```

This also gets rid of the `escapedField` var (although that could have been removed anyway). Notice that all it does is possibly cause an extra copy. If you have to escape the string, `escapedField` is copied into once, and then it's overwritten at the end.

src/common/CSV.cpp
```diff
@@ -44,6 +42,121 @@
...
+        escapedField = escaped.str();
+    }
+
+    return escapedField;
+}
+
+// Function to unescape a single CSV field
+std::string unescape_csv_field(const std::string& field) {
+    // If the field starts and ends with a quote, it is escaped
+    if (field.size() >= 2 && field.front() == '"' && field.back() == '"') {
```
jblee123: Another instance where the entire function is in a block. Favor:
```
    bool is_quoted_string = field.size() >= 2 && field.front() == '"' && field.back() == '"';
    if (!is_quoted_string) {
        return field;
    }

    // logic to un-escape
    return unescaped.str();
```
Could you do that without the `is_quoted_string` intermediate var? Yes. It wouldn't look terrible. However, it's nice to have because:
1. It nicely labels what you're actually checking for, and
2. That's a much more intuitive test than the "not-a-quoted-string" you'd have to put in the `if` statement -- you'd have to use DeMorgan's to invert the test, and the inverted test would not be as intuitive.

src/common/CSV.cpp
```diff
@@ -44,6 +42,121 @@
...
+        // Process the part within the quotes
+        for (size_t i = 1; i < field.size() - 1; ++i) {
+            char c = field[i];
+            if (isQuotePair) {
+                if (c == '"') {  // This is the double quote -> unescape it
+                    unescaped << '"';
+                    isQuotePair = false;
+                } else {
+                    // Invalid situation (should not happen in a valid CSV): ignore
+                    throw std::runtime_error("Invalid CSV: unexpected character after quote");
```
jblee123: The comment says "ignore", but you throw an exception, which are two very different things.

Just my opinion, but exceptions suck. Do you wrap the calling code in try/catch elsewhere? Do you expect clients to? How would they know they needed to?

The error message -- it tells what went wrong, but put yourself in the place of a user. Here's this 10,000 line long CSV file, and all they get is "unexpected character after quote". Now what? Help them out. At least give them the bad text. Maybe even the line number. Don't make your users hate you.

src/common/CSV.cpp
```diff
@@ -44,6 +42,121 @@
 
 namespace scrimmage {
 
+// Function to escape special characters in a CSV field
+std::string escape_csv_field(const std::string& field) {
```
jblee123: Since we've got `std::string_view` now in c++20, consider using those as params where you'd usually use a `const std::string&`. It avoid unnecessary copies if `char*`s are passed in, but you still get all the nice functions of a `std::string`.

src/common/CSV.cpp
```diff
@@ -44,6 +42,121 @@
...
+        return unescaped.str();
+    }
+
+    // If there are no surrounding quotes, return the field as is
+    return field;
+}
+
+// Function to split a CSV string by some token while respecting escape sequences of
+// 2 double quotes """"
+std::vector<std::string> split_csv_string(const std::string& row, const char& token) {
```
jblee123: `const char& token`... a whole reference to a `char` -- so passing 8 bytes instead of 1, and requiring following the reference back to the original every time. Use `char`.

src/common/CSV.cpp
```diff
@@ -44,6 +42,121 @@
...
+                // Append any other character inside quotes
+                // This is where newlines and commas can appear
+                // as valid text inside a cell
+                currentField << c;
+            }
+        } else {
+            if (c == token) {
+                // token outside quotes: field delimiter
+                fields.push_back(currentField.str());  // Add the current field to the list
+                currentField.str("");                  // Reset for the next field
```
jblee123, line 140:
```
                currentField.str("");                  // Reset for the next field
                currentField.clear();                  // Reset for the next field
```
is redundant.

src/common/CSV.cpp
```diff
@@ -44,6 +42,121 @@
...
+    bool insideQuotes = false;        // Tracks if we're inside a quoted field
+
+    for (size_t i = 0; i < row.size(); ++i) {
+        char c = row[i];
+
+        if (insideQuotes) {
+            if (c == '"') {
+                // If we see a double-quote while inside quotes, check for escape (i.e., `""`)
+                if (i + 1 < row.size() && row[i + 1] == '"') {
+                    currentField << c;
```
jblee123:
```'
                    currentField << c;
                    currentField << c;
```
This works, but please no... it looks like an accidentally repeated line at first, and you have to think more about surrounding context to figure out that it's not. Just do:
```
                    currentField << "\"\"";
```
It's one call, and it tells you exactly what's going in as opposed to relying on you to figure it out because you know `c` is `'='`.

jblee123 (commented):
I don't think I saw anything that wouldn't work, but there were some stylistic choices.

PR #631, opened 2026-03-31
adding an arm pipeline build and supporting docs

EthanMBoos:


development-docs/CONTAINERIZED_DEVELOPMENT.md
```diff
@@ -37,13 +47,24 @@ This automatically configures clangd intellisense and X11 forwarding (Linux).
...
+### Auto-detect architecture (recommended)
+
 ```bash
 docker run -it ghcr.io/gtri/scrimmage-24.04:latest bash
 source /root/scrimmage/build/install/etc/scrimmage/env/scrimmage-setenv
 scrimmage /root/scrimmage/missions/straight-no-gui.xml
 ```
 
-For GUI support:
+### Specify architecture explicitly
```
wfsyre: I think you can remove this header from the docs. The --platform flag is a docker feature so they should be able to find documentation for that in the docker docs

wfsyre (dismissed):

wfsyre (approved):

PR #632, opened 2026-04-01
Add robust logging framework

EthanMBoos:
Going to add some unit tests to make sure it looks good, so stay posted.


src/log/Logger.cpp
```diff
@@ -27,24 +27,67 @@
...
 #include <iomanip>
 #include <iostream>
 #include <sstream>
 
 namespace scrimmage {
 
 Logger::~Logger() {
-    close();
+    std::lock_guard<std::mutex> lock(mutex_);
+    // If init_dir was never called and we have pending messages, write to fallback
```
wfsyre: Can we get a warning or something here indicating the user won't find logs in the spot they were expecting (or where to find them if they never set it)

src/simcontrol/SimControl.cpp
```diff
@@ -148,9 +148,12 @@ void SimControl::send_terrain() {
 bool SimControl::setup_logging() {
     limited_verbosity_ = mp_->output_required();
 
-    // Setup the log directory if it is required
+    // Always create the log directory and redirect the text logger to it.
+    mp_->create_log_dir();
```
wfsyre: I see why you did this, but its still a little upsetting that if a user wants no logging, the mission parser will still generate a folder on their system and then just not put logs in it. It would be nice if we could run with 0 footprint if asked to

src/parse/ParseUtils.cpp
```diff
@@ -98,6 +98,7 @@ bool find_terrain_files(
             return true;
         }
     }
+    LOG_WARN("Terrain '" << terrain_name << "' not found or files missing (polydata/texture)");
```
wfsyre: Lets add the path that we expect terrain files to be in to this warning (SCRIMMAGE_DATA_PATH)

src/parse/ParseUtils.cpp
```diff
@@ -365,20 +366,26 @@ bool set_pid_gains(sc::PID& pid, std::string str, bool is_angle) {
...
+            double d = std::stod(str_vals[2]);
+            pid.set_parameters(p, i, d);
+
+            if (is_angle) {
+                double i_lim = sc::Angles::deg2rad(std::stod(str_vals[3]));
+                pid.set_integral_band(i_lim);
+                pid.set_is_angle(true);
+            } else {
+                double i_lim = std::stod(str_vals[3]);
+                pid.set_integral_band(i_lim);
```
wfsyre, line 389: Lets set pid.is_angle(false) explicitly here. Not your code really but while you're here we can make this 1% more readable

src/simcontrol/SimControl.cpp
```diff
@@ -1573,27 +1599,35 @@ void SimControl::worker() {
...
-                success = std::all_of(controllers.begin(), controllers.end(), run);
-            } else if (task_type == Task::Type::MOTION) {
-                success = ent->motion()->step(temp_t, temp_dt);
-            } else if (task_type == Task::Type::SENSOR) {
-                auto sensors = ent->sensors() | ba::map_values;
-                br::for_each(sensors, run_callbacks);
-                auto run = [&](auto& s) { return s->step_loop_timer(temp_dt) ? s->step() : true; };
-                success = std::all_of(sensors.begin(), sensors.end(), run);
+            try {
+                if (task_type == Task::Type::AUTONOMY) {
```
wfsyre, line 1636: This block that uses step_loop_timer is one of those things that's been incorrect for awhile. t and dt should be updated based on what was last sent to the plugin. So first loop should be t and dt/num_loops and next loop should be t + 1 * (dt / num_loops). This keeps the plugins agnostic to the fact that they are looping and instead just feeds them normal t and dt as if they were just running with a higher rate. Maybe a fix for another commit, but if you see and easy way of doing this here then I think its a good time

src/plugins/autonomy/Straight/Straight.cpp
```diff
@@ -171,11 +171,9 @@ void Straight::init(std::map<std::string, std::string>& params) {
                 } else {
                     // other image types are int 0-255.
                     if (a.img.channels() == 4) {
-                        cout << "image channels: " << a.img.channels() << endl;
-                        cout << "Warning: Old AirSim Linux Asset Environments have 4 channels. "
-                                "Color images will not display correctly."
-                             << endl;
-                        cout << "Warning: Use Asset Environment versions Linux-v1.3.1+." << endl;
+                        LOG_INFO("image channels: " << a.img.channels());
+                        LOG_WARN("Warning: Old AirSim Linux Asset Environments have 4 channels. Color images will not display correctly.");
```
wfsyre: I think your logger already prepends the Warn so we can remove "Warning" from these ones

wfsyre (changes requested):
No reviews on your most recent commit because its WIP

wfsyre (approved):

PR #637, opened 2026-04-05
lldb configuration added

EthanMBoos:
adding vscode configs as a baseline then going to backout some neovim ones. neovim debuggers are a little trickier

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,6 @@ tags*
 **.swp
 **.DS_Store
 **.ccls*
-.vscode
 env
 .idea*
 python/scrimmage/bindings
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -36,7 +36,6 @@ tags*
 **.swp
 **.DS_Store
 **.ccls*
-.vscode
```
wfsyre, line 39: Not sure if this is possible or what the .vscode folder usually contains, but consider whitelisting just the files you added to .vscode and not allowing just any file from that folder in to the git tree

wfsyre (dismissed):

wfsyre (approved):

PR #638, opened 2026-04-06
Cli positional overrides

EthanMBoos:
Now for example you can run a mission like `scrimmage enable_gui:=false missions/straight.xml`

Much nicer syntax than `scrimmage -o "enable_gui=false" missions/straight.xml`

This is convenient for the ci/cd work I am doing where quotes often cause issues and need to be escaped.

### DIFF
--- a/share/scrimmage/main.cpp
+++ b/share/scrimmage/main.cpp
@@ -35,6 +35,7 @@
 #include <iostream>
 #include <memory>
 #include <ostream>
+#include <sstream>
 #include <string>
 #include <unordered_set>
 
@@ -95,7 +96,8 @@ int main(int argc, char* argv[]) {
 
     bool seed_set = false;
     std::string seed = "";
-    std::string overrides = "";
+    std::ostringstream overrides;
+    bool first_override = true;
 
     int opt;
     while ((opt = getopt(argc, argv, "t:j:s:")) != -1) {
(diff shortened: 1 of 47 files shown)


share/scrimmage/main.cpp
```diff
@@ -98,7 +98,7 @@ int main(int argc, char* argv[]) {
     std::string overrides = "";
```
wfsyre: Efficiency wise, if you are appending to a string a lot you should make it a std::ostringstream and use 
old_str << append_value;

share/scrimmage/main.cpp
```diff
@@ -126,7 +123,29 @@ int main(int argc, char* argv[]) {
...
+    // Parse positional arguments: args with ":=" are overrides, first without is mission file
+    std::string mission_file;
+    for (int i = optind; i < argc; ++i) {
+        std::string arg(argv[i]);
+        if (arg.find(":=") != std::string::npos) {
+            // This is an override
+            if (!overrides.empty()) {
+                overrides += ",";
+            }
+            overrides += arg;
```
wfsyre: Can we normalize the := into just a = here so that mission parse doesn't need to know about this?

share/scrimmage/main.cpp
wfsyre: Then just old_str.str() to convert back to std::string and pass it along

wfsyre (changes requested):

wfsyre (approved):

PR #639, opened 2026-04-07
mission syntax validation

EthanMBoos:


src/parse/MissionValidation.cpp
```diff
@@ -0,0 +1,319 @@
...
+        }
+        std::string plugin_name = filename.substr(0, filename.size() - 4);
+
+        // Extract type from path: .../plugins/<type>/<PluginName>/...
+        for (const std::string& full_path : paths) {
+            size_t plugins_pos = full_path.find("/plugins/");
+            if (plugins_pos == std::string::npos) continue;
+
+            // Get the directory after "plugins/"
+            size_t type_start = plugins_pos + 9;  // len("/plugins/")
```
wfsyre: Lets make this a string variable "/plugins/" and pass that into the find as well as len(plugin_var) for the type_start traverse to the end of the string

src/parse/MissionValidation.cpp
```diff
@@ -0,0 +1,319 @@
...
+            size_t type_end = full_path.find('/', type_start);
```
wfsyre: Sometimes we get paths like plugins//type//name. The extra slashes don't matter to the OS and they shouldn't matter to us either. Lets use a dedicated filesystem iterator like https://en.cppreference.com/w/cpp/filesystem/directory_iterator.html

src/parse/MissionValidation.cpp
```diff
@@ -0,0 +1,319 @@
...
+            size_t cost = (s1[i - 1] == s2[j - 1]) ? 0 : 1;
+            dp[i][j] = std::min({
+                dp[i - 1][j] + 1,
+                dp[i][j - 1] + 1,
+                dp[i - 1][j - 1] + cost
+            });
+        }
+    }
+    return dp[m][n];
+}
```
wfsyre, line 276: Wow dynamic programming in the flesh, what a treat

src/parse/MissionValidation.cpp
```diff
@@ -0,0 +1,319 @@
...
+        const auto& params = ent_kv.second;
+        std::string context = "entity block " + std::to_string(entity_id);
+
+        auto it_motion = params.find("motion_model");
+        if (it_motion != params.end() && !it_motion->second.empty()) {
+            plugin_names["motion_model"].push_back({it_motion->second, context});
+        }
+
+        // autonomy0, autonomy1, ...
+        for (int i = 0; ; ++i) {
```
wfsyre: If we're doing validation work on these items can we also store them with the plugin's real name instead of autonomy0? Might require you to go a bit higher up in the parsing structure but a struct of some kind that holds: Plugin name, plugin order, and any other metadata would be a good thing to store for plugins in general. We have another system that parses this and edits params and right now plugin params are changes by adding a param to "autonomy0, 1, etc" which is not ideal since the ordering of plugins affects that name. If we can standardize by a guaranteed unique identifier then we're one step closer to breaking out of that paradigm

src/parse/MissionValidation.cpp
```diff
@@ -0,0 +1,319 @@
...
+    for (const auto& kv : entity_plugins) {
+        const std::string& plugin_type = kv.first;
+        const auto& available = result.available_plugins[plugin_type];
+
+        for (const auto& name_ctx : kv.second) {
+            const std::string& plugin_name = name_ctx.first;
+            const std::string& context = name_ctx.second;
+
+            std::string cache_key = plugin_type + ":" + plugin_name;
+            if (validated_plugins_.count(cache_key) > 0 || plugin_name.empty()) {
```
wfsyre: Does this just ignore duplicate plugins silently? Or are you expecting to see the same plugin multiple times? We have a problem right now where if we see duplicate plugins in the path we just pick one randomly. I would love to do something intelligent instead like using the one that's closest in path to the mission file we're running.

wfsyre (changes requested):

src/parse/MissionValidation.cpp
EthanMBoos: Funny algo name too - Levenshtein

EthanMBoos:
I removed the ability to override plugins at runtime through GenerateEntity. If you try to set autonomy or controller via entity_param, it'll warn you and ignore it now. I didn't see any usage in the Scrimmage repo — the only real usage was for things like color, not plugin names. The docs made it look like it worked, but it was dead code adding complexity. If you need different plugin setups, just make separate entity templates.

```xml
<!-- Fast version -->
<entity tag="fast_uav">
  <count>0</count>
  <autonomy speed="30">Straight</autonomy>
  <controller>SimpleAircraftControllerPID</controller>
  ...
</entity>

<!-- Slow version -->
<entity tag="slow_uav">
  <count>0</count>
  <autonomy speed="10">Straight</autonomy>
  <controller>SimpleAircraftControllerPID</controller>
  ...
</entity>
```

Set count=0 so they don't spawn automatically
```c++
msg->data.set_entity_tag("fast_uav");  // spawns the fast one
// or
msg->data.set_entity_tag("slow_uav");  // spawns the slow one
```

src/parse/MissionValidation.cpp
EthanMBoos: You sneaky man. I thought this was a simple front-end "check some syntax" easy-peasy quality of life improvement.

I went off the deep end and just did the whole refactor. The higher up I went, the more there was to change... all hardcoded autonomy0,1 control0,1 ...

EthanMBoos:
- Existing plugins that used the old runtime override path will need to migrate.
- Reintroduced runtime plugin parameter overrides for `GenerateEntity` via `plugin_override`, targeting plugins by `plugin_name` or `plugin_index` instead of free-form strings like `autonomy0`. Unknown runtime keys are now validated by default; set `strict_runtime_plugin_params=false` only for loose override behavior.

src/entity/RuntimePluginOverrides.cpp
```diff
@@ -0,0 +1,306 @@
...
+                     << "'. Expected one of autonomy, controller, motion_model, sensor.");
+            return false;
+        }
+
+        // Resolve plugin_name to plugin_index if specified
+        int resolved_index = static_cast<int>(proto_override.plugin_index());
+        if (proto_override.has_plugin_name()) {
+            resolved_index = resolve_plugin_name_to_index(
+                mp, ent_desc_id, proto_override.plugin_type(), proto_override.plugin_name());
+            if (resolved_index == -1) {
```
wfsyre: Lets add a constexpr int kNoPluginNameFound = -1 into your anonymous namespace up top. That way its immediately obvious what's going on in both this context and the function that returns -1 instead of an index.

src/entity/RuntimePluginOverrides.cpp
```diff
@@ -0,0 +1,306 @@
...
+                LOG_ERROR("GenerateEntity: Plugin name '" << proto_override.plugin_name()
+                         << "' not found in " << proto_override.plugin_type()
+                         << " plugins for entity block " << ent_desc_id << ".");
+                return false;
+            }
+            if (resolved_index == -2) {
```
wfsyre: Same here

msgs/Event.proto
```diff
@@ -40,9 +40,17 @@ message KeyValuePair {
...
+        string plugin_type = 1;
+        uint32 plugin_index = 2;
+        repeated KeyValuePair params = 3;
+        optional string plugin_name = 4; // name will resolve to a plugin_index
+}
+
 message GenerateEntity {
         string entity_tag = 1;
         scrimmage_proto.State state = 2;
         repeated KeyValuePair entity_param = 3;
```
wfsyre: This is what your system is replacing right? If so, then break it and remove it. I don't need this to be backwards compatible if the piece that actually populates the runtime parameters has been disabled. Give me a compiler error so I know its wrong.

wfsyre (changes requested):

msgs/Event.proto
EthanMBoos: removed

src/entity/RuntimePluginOverrides.cpp
EthanMBoos: added constexpr

src/entity/RuntimePluginOverrides.cpp
EthanMBoos: added constexpr

PR #640, opened 2026-04-08
edits to make install build process work for external repos

EthanMBoos:


scripts/templates/cmake-project/CMakeLists.txt
```diff
@@ -78,6 +78,7 @@ set(PROTO_INCLUDE_DIR ${PROJECT_BINARY_DIR}/src/proto)
 ###############################################################################
 find_package(scrimmage REQUIRED)
 
+option(EXTERNAL "Mark as external project" ON)
```
wfsyre, line 81: Can we get some documentation here on what EXTERNAL means to the user that discovers this cmake setting?

wfsyre (changes requested):

PR #641, opened 2026-05-01
added O(n) spatial collision checker.

EthanMBoos:
- drop in replacement for O(n^2) SimpleCollision

### DIFF
--- a/src/plugins/interaction/SpatialCollision/SpatialCollision.cpp
+++ b/src/plugins/interaction/SpatialCollision/SpatialCollision.cpp
@@ -1,0 +1,227 @@
+/*!
+ * @file
+ *
+ * @section LICENSE
+ *
+ * Copyright (C) 2017 by the Georgia Tech Research Institute (GTRI)
+ *
+ * This file is part of SCRIMMAGE.
+ *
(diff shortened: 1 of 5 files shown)


EthanMBoos:
I ran a few sample missions here with different numbers of agents. Looks to save 30%-40% of time depending on the mission. Could be good for bigger bulk sim runs.

I also verified # of hits was the same on each run, but this was on the toy straight.xml mission and I am not confident that 100% of the time # of SpatialCollisions = SimpleCollisions. Needs some side by side testing.

src/plugins/interaction/SpatialCollision/SpatialCollision.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    bool operator==(const CellIndex& other) const {
+        return x == other.x && y == other.y && z == other.z;
+    }
+};
+
+struct CellIndexHash {
+    std::size_t operator()(const CellIndex& cell) const {
+        std::size_t seed = std::hash<std::int64_t>{}(cell.x);
+        seed ^= std::hash<std::int64_t>{}(cell.y) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```
wfsyre, line 87: What is this magic number here?

